PR #5, opened 2017-10-29
MO: Introduced Cross Domain tracking

manfield:
In some circumstances cross-domain tracking could be useful.
This pull request introduces cross-domain tracking if Multistore is enabled and if option is active.
Moreover, the GA script is moved to a template, thus may be overridden easily by users if needed.

### DIFF
--- a/ps_googleanalytics.php
+++ b/ps_googleanalytics.php
@@ -118,6 +118,9 @@ class Ps_Googleanalytics extends Module
 
     public function displayForm()
     {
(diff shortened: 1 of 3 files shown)


sharkooon:
Prestashop please learn from the community and keep your repositories alive! With this pullrequest it also works for prestashop 1.7.2! so why does it take so long??

Thank you manfield, its working!

manfield:
You're welcome!

PR #14, opened 2018-09-28
Only cast as array when a product is returned

Quetzacoalt91:
Product content should be casted as an array only if we have an Product instance.

With the LazyArrays being introduced, casting these variables as arrays breaks the module execution. See details on https://github.com/PrestaShop/PrestaShop/issues/10742


marionf:
I can't change the label, but it's QA approved :+1:

eternoendless:
Thank you @Quetzacoalt91

PR #15, opened 2018-10-10
Fix wordings without domain

eternoendless:


PierreRambaud:
Thanks @eternoendless

PR #23, opened 2018-12-27
Fopen flags

MathiasReker:


PierreRambaud:
This one is not needed. Thanks @MathiasReker

PR #37, opened 2019-11-04
Add pageview

david-piatek:


sarahdib:
@david-piatek 
Can you give us test case ?

Thank you

PierreRambaud:
Thanks @david-piatek

jf-viguier:
page view is fired twice on a 1.7.6.2, I've found another pageview event here :
https://github.com/PrestaShop/ps_googleanalytics/blob/a41f865a66e6ba3426db1f48121e1b7c0d034f0b/ps_googleanalytics.php#L747
Are-you sure of this pr ?

david-piatek:
I don't understand the role of js_state, but it sometimes blocks execution of ga.
That’s why I fixed it as little as possible to the existing code ...
Does this have side effects?

Slews:
@david-piatek the update seems to cause the "pageview" send called twice.

Can you check on the project you had the execution of ga blocked whereas the ps_googleanalytics is attached to the proper list of hooks it should ?

The reason I am asking is that I saw some of my project did have the twice call but some didn't. And searching why, it appeared that some projects were using a theme with a bad theme.yml occuring the module to be detached of many hooks.
Particuliarly, the pageview call seems to need the module to be attached to displayFooter.

dali-rajab:
@PierreRambaud  @david-piatek @khouloudbelguith @sarahdib 

Since we want to track every page view, I suggest to put back `ga('send', 'pageview');` in **ps_googleanalytics.tpl** to be hooked in every front-office page header, thus, we don't have to deal anymore with `ga('send', 'pageview');`, so we should :

- remove every `$this->js_state = ...`
- remove `protected $js_state = 0;`
- remove `     if (($this->js_state) != 1 && ($backoffice == 0)) {$runjs_code .= '<script type="text/javascript">ga(\'send\', \'pageview\');</script>'; }`
- remove `ga('send', 'pageview');` from **GoogleAnalyticActionLib.js**

I tested it, it works like a sharm. Should i make a pull request for these changes ??
Thank you for your suggestions

HartLarsson:
> * ga('send', 'pageview')

i did the same and working removing the double pageview sent!

I never understand the idea of $this->js_state =X
what is needed for?

camlafit:
Hello

Could be possible to revert this patch ? Or complete it ? Actually events are fired twice and tracking data are false

Thanks a lot :)

PR #43, opened 2020-02-21
Add tracking of cancelled orders

Hlavtox:
This PR enables tracking of cancelled orders, which could greatly affect the statistics. The refund functionality is already provided by the enhanced ecommerce library.

**Changes**
- The functionality hooks to actionOrderStatusPostUpdate.
- The functionality adds 1 column to ps_ganalytics to prevent multiple refunds for one order.
- Bumps module version to 3.3.0 to run the upgrade script needed.

**How it works**
- After you change the order status, the module checks, if the new status is CANCELLED.
- If so, it checks the database, if this order was not cancelled before - to prevent multiple refunds, which Google Analytics unfortunately allows.
- If no refund information found in database, adds the refund action by MBG.refundByOrderId.

**How to test**
1. Install Google Tag Assistant (in Chrome).
2. Install this module, set it up.
3. Make an order, go to it's administration.
4. Enable recording in Google Tag Assistant.
5. Change order status to cancelled.
6. After page refresh, check Google Tag Assistant and see the Refund event being sent.
(After some hours, the transaction will disappear from Google Analytics.)
7. Change order status to "Processing", then, change it back to "Cancelled".
8. No refund event should be sent.

### DIFF
--- a/ps_googleanalytics.php
+++ b/ps_googleanalytics.php
@@ -77,6 +77,7 @@ class Ps_Googleanalytics extends Module
             $this->registerHook('actionProductCancel') &&
             $this->registerHook('actionCartSave') &&
             $this->registerHook('displayBackOfficeHeader') &&
+	    $this->registerHook('hookactionOrderStatusPostUpdate') &&
             $this->registerHook('actionCarrierProcess')
         ) {
             return $this->createTables();
@@ -844,6 +845,20 @@ class Ps_Googleanalytics extends Module
             return $js.$this->hookdisplayHeader(null, true).$this->_runJs($ga_scripts, 1);
         } else {
             return $js;
+        }
+    }
+
+    /**
+     * Hook called after order status change
+     * Used to cancel order after cancelling it
+     */
+    public function hookactionOrderStatusPostUpdate($params)
+    {
+        $order_status = $params['newOrderStatus']->id;
+
+        // If we have the order and new order status in CANCELED
+        if (!empty($params['id_order']) && !empty($order_status) && $order_status == (int) Configuration::get('PS_OS_CANCELED')) {
+            $this->context->cookie->ga_admin_refund = $ga_scripts . 'MBG.refundByOrderId(' . json_encode(['id' => $params['id_order']]) . ');';
         }
     }
 
--- a/ps_googleanalytics.php
+++ b/ps_googleanalytics.php
@@ -873,7 +873,7 @@ class Ps_Googleanalytics extends Module
             $this->context->cookie->ga_admin_refund = $ga_scripts . 'MBG.refundByOrderId(' . json_encode(['id' => $params['id_order']]) . ');';
             Db::getInstance()->execute(
                 'UPDATE `'._DB_PREFIX_.'ganalytics` SET refund_sent = 1 WHERE id_order = '. (int) $params['id_order']
-	    );
+            );
         }
 
     }
--- a/ps_googleanalytics.php
+++ b/ps_googleanalytics.php
@@ -77,7 +77,7 @@ class Ps_Googleanalytics extends Module
             $this->registerHook('actionProductCancel') &&
             $this->registerHook('actionCartSave') &&
             $this->registerHook('displayBackOfficeHeader') &&
-	    $this->registerHook('hookactionOrderStatusPostUpdate') &&
+            $this->registerHook('hookactionOrderStatusPostUpdate') &&
             $this->registerHook('actionCarrierProcess')
         ) {
             return $this->createTables();
@@ -852,12 +852,10 @@ class Ps_Googleanalytics extends Module
      * Hook called after order status change
      * Used to cancel order after cancelling it
      */
-    public function hookactionOrderStatusPostUpdate($params)
+    public function hookactionOrderStatusPostUpdate(array $params)
     {
-        $order_status = $params['newOrderStatus']->id;
-
-        // If we have the order and new order status in CANCELED
-        if (!empty($params['id_order']) && !empty($order_status) && $order_status == (int) Configuration::get('PS_OS_CANCELED')) {
+        // If we have the order AND the new order status being set AND the order status is changing to CANCELLED
+        if (!empty($params['id_order']) && !empty($params['newOrderStatus']->id) && $params['newOrderStatus']->id == (int) Configuration::get('PS_OS_CANCELED')) {
             $this->context->cookie->ga_admin_refund = $ga_scripts . 'MBG.refundByOrderId(' . json_encode(['id' => $params['id_order']]) . ');';
         }
     }
--- a/ps_googleanalytics.php
+++ b/ps_googleanalytics.php
@@ -77,7 +77,7 @@ class Ps_Googleanalytics extends Module
             $this->registerHook('actionProductCancel') &&
             $this->registerHook('actionCartSave') &&
             $this->registerHook('displayBackOfficeHeader') &&
-            $this->registerHook('hookactionOrderStatusPostUpdate') &&
+            $this->registerHook('actionOrderStatusPostUpdate') &&
             $this->registerHook('actionCarrierProcess')
         ) {
             return $this->createTables();
@@ -853,21 +853,29 @@ class Ps_Googleanalytics extends Module
      * Hook called after order status change
      * Used to cancel order after cancelling it
      */
-    public function hookactionOrderStatusPostUpdate(array $params)
+    public function hookActionOrderStatusPostUpdate(array $params)
     {
         // If we have the order AND the new order status being set AND the order status is changing to CANCELLED
-        if (!empty($params['id_order']) && !empty($params['newOrderStatus']->id) && $params['newOrderStatus']->id == (int) Configuration::get('PS_OS_CANCELED')) {
-        
-            // We check if the refund was already sent to Google Analytics
-            $ga_refund_sent = Db::getInstance()->getValue('SELECT id_order FROM `' . _DB_PREFIX_ . 'ganalytics` WHERE id_order = ' . (int) $params['id_order'] . ' && refund_sent = 1');
-            
-            if ($ga_refund_sent === false) {
-                // We refund it and set the "sent" flag to true
-                $this->context->cookie->ga_admin_refund = $ga_scripts . 'MBG.refundByOrderId(' . json_encode(['id' => $params['id_order']]) . ');';
-                Db::getInstance()->execute('UPDATE `'._DB_PREFIX_.'ganalytics` SET refund_sent = 1 WHERE id_order = '. $params['id_order']);
-            }
-		
+        if (empty($params['id_order'])
+            || empty($params['newOrderStatus']->id)
+            || (int) $params['newOrderStatus']->id !== (int) Configuration::get('PS_OS_CANCELED')
+        ) {
+            return;
         }
+	            
+        // We check if the refund was already sent to Google Analytics
+        $gaRefundSent = Db::getInstance()->getValue(
+            'SELECT id_order FROM `' . _DB_PREFIX_ . 'ganalytics` WHERE id_order = ' . (int) $params['id_order'] . ' AND refund_sent = 1'
+        );
+            
+        if ($gaRefundSent === false) {
+            // We refund it and set the "sent" flag to true
+            $this->context->cookie->ga_admin_refund = $ga_scripts . 'MBG.refundByOrderId(' . json_encode(['id' => $params['id_order']]) . ');';
+            Db::getInstance()->execute(
+                'UPDATE `'._DB_PREFIX_.'ganalytics` SET refund_sent = 1 WHERE id_order = '. (int) $params['id_order']
+	    );
+        }
+
     }
 
     /**
--- a/ps_googleanalytics.php
+++ b/ps_googleanalytics.php
@@ -46,7 +46,7 @@ class Ps_Googleanalytics extends Module
     {
         $this->name = 'ps_googleanalytics';
         $this->tab = 'analytics_stats';
-        $this->version = '3.2.0';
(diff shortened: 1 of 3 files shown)


ps_googleanalytics.php
```diff
@@ -76,6 +76,7 @@ public function install()
             $this->registerHook('actionProductCancel') &&
             $this->registerHook('actionCartSave') &&
             $this->registerHook('displayBackOfficeHeader') &&
+	    $this->registerHook('hookactionOrderStatusPostUpdate') &&
```
PierreRambaud: Wrong indentation
```suggestion
	    $this->registerHook('hookActionOrderStatusPostUpdate') &&
```

ps_googleanalytics.php
```diff
@@ -841,6 +842,20 @@ public function hookdisplayBackOfficeHeader()
         }
     }
 
+    /**
+     * Hook called after order status change
+     * Used to cancel order after cancelling it
+     */
+    public function hookactionOrderStatusPostUpdate($params)
```
PierreRambaud:
```suggestion
    public function hookActionOrderStatusPostUpdate($params)
```

ps_googleanalytics.php
```diff
@@ -841,6 +842,20 @@ public function hookdisplayBackOfficeHeader()
...
+    {
+        $order_status = $params['newOrderStatus']->id;
```
PierreRambaud: Make sure this accessor exists before using it please.

Hlavtox:
@PierreRambaud Hi Pierre, I added a test scenario to the PR description. 🍺

Hlavtox:
@PierreRambaud Hi Pierre, I have a question. I slightly modified the code to prevent doubled refunds.

Would you have time to help me with running the upgrade script after upgrade? I don't know the upgrade logic.

`ALTER TABLE `ps_ganalytics` ADD `refund_sent` tinyint(1) NULL AFTER `sent`;`

**The code**

  ```
      if (we have all data) {

            // we check if the refund was already sent to ga
            $ga_refund_sent = Db::getInstance()->getValue('SELECT id_order FROM `' . _DB_PREFIX_ . 'ganalytics` WHERE id_order = ' . (int) $params['id_order'] . ' && refund_sent = 1');

            if ($ga_refund_sent === false) {
                // We refund it and set the "sent" flag to true
                $this->context->cookie->ga_admin_refund = $ga_scripts . 'MBG.refundByOrderId(' . json_encode(['id' => $params['id_order']]) . ');';
                Db::getInstance()->execute('UPDATE `'._DB_PREFIX_.'ganalytics` SET refund_sent = 1 WHERE id_order = '. $params['id_order']);
            }
            
        }
```

**Comment by PierreRambaud** [comment]
Feel free to get inspired by https://github.com/PrestaShop/ps_facetedsearch/tree/master/upgrade
https://devdocs.prestashop.com/1.7/modules/creation/enabling-auto-update/

**Comment by Hlavtox** [comment]
@PierreRambaud Thanks, seems easy enough!

One question though. How should I approach this? Should I propose a change in the module version and name the corresponding upgrade file the same way?

**Comment by PierreRambaud** [comment]
You can upgrade the version to `3.3.0` if you want :)

**Comment by Hlavtox** [comment]
@PierreRambaud Done boss, can you check it? 🚀 🍻

File: ps_googleanalytics.php [hunk 4]
```diff
@@ -76,6 +76,7 @@ public function install()
             $this->registerHook('actionProductCancel') &&
             $this->registerHook('actionCartSave') &&
             $this->registerHook('displayBackOfficeHeader') &&
+            $this->registerHook('hookactionOrderStatusPostUpdate') &&
```
PierreRambaud, line 79:
```suggestion
            $this->registerHook('actionOrderStatusPostUpdate') &&
```

ps_googleanalytics.php
```diff
@@ -841,6 +843,27 @@ public function hookdisplayBackOfficeHeader()
         }
     }
 
+    /**
+     * Hook called after order status change
+     * Used to cancel order after cancelling it
+     */
+    public function hookactionOrderStatusPostUpdate(array $params)
```
PierreRambaud, line 850:
```suggestion
    public function hookActionOrderStatusPostUpdate(array $params)
```

ps_googleanalytics.php
```diff
@@ -841,6 +843,27 @@ public function hookdisplayBackOfficeHeader()
...
+            $ga_refund_sent = Db::getInstance()->getValue('SELECT id_order FROM `' . _DB_PREFIX_ . 'ganalytics` WHERE id_order = ' . (int) $params['id_order'] . ' && refund_sent = 1');
+            
+            if ($ga_refund_sent === false) {
+                // We refund it and set the "sent" flag to true
+                $this->context->cookie->ga_admin_refund = $ga_scripts . 'MBG.refundByOrderId(' . json_encode(['id' => $params['id_order']]) . ');';
+                Db::getInstance()->execute('UPDATE `'._DB_PREFIX_.'ganalytics` SET refund_sent = 1 WHERE id_order = '. $params['id_order']);
+            }
+		
+        }
+    }
```
PierreRambaud, line 865:
```suggestion
    {
    // If we have the order AND the new order status being set AND the order status is changing to CANCELLED
    if (empty($params['id_order'])
        || !empty($params['newOrderStatus']->id)
        || (int) $params['newOrderStatus']->id !== (int) Configuration::get('PS_OS_CANCELED')
    ) {
        return;
    }
        
    // We check if the refund was already sent to Google Analytics
    $gaRefundSent = Db::getInstance()->getValue(
        'SELECT id_order FROM `' . _DB_PREFIX_ . 'ganalytics` WHERE id_order = ' . (int) $params['id_order'] . ' AND refund_sent = 1'
    );
            
    if ($gaRefundSent === false) {
        // We refund it and set the "sent" flag to true
        $this->context->cookie->ga_admin_refund = $ga_scripts . 'MBG.refundByOrderId(' . json_encode(['id' => $params['id_order']]) . ');';
        Db::getInstance()->execute('UPDATE `'._DB_PREFIX_.'ganalytics` SET refund_sent = 1 WHERE id_order = '. (int) $params['id_order']);
    }
```

Hlavtox:
@PierreRambaud Sorry, I did not see the misspelled hook during install. :-(

I will apply the change it, test it (including the upgrade script) and let you know.

Hlavtox:
@PierreRambaud Hi, I double checked everything and I think it should be good to go. 

- Install/uninstall works fine and hooks where it should.
- Upgrade works fine and updates the table correctly.
- Functionality works fine.

Quetzacoalt91:
Rebased pull-request

ps_googleanalytics.php
```diff
@@ -847,6 +849,35 @@ public function hookdisplayBackOfficeHeader()
...
+        }
+	            
+        // We check if the refund was already sent to Google Analytics
+        $gaRefundSent = Db::getInstance()->getValue(
+            'SELECT id_order FROM `' . _DB_PREFIX_ . 'ganalytics` WHERE id_order = ' . (int) $params['id_order'] . ' AND refund_sent = 1'
+        );
+            
+        if ($gaRefundSent === false) {
+            // We refund it and set the "sent" flag to true
+            $this->context->cookie->ga_admin_refund = $ga_scripts . 'MBG.refundByOrderId(' . json_encode(['id' => $params['id_order']]) . ');';
```
SimonGrn, line 873: Where do you get the variable `$ga_scripts` ?
I tried your PR in a production environment with my own personal GA ID and I got this error with debug mode on :
image

Moreover, GA is complaining about a "No HTTP response detected" : 
image

Hlavtox: Hi @SimonGrn, my mistake, it was an empty unneeded variable.

It's complaining because no view is sent to GA, only events are sent. Try the cancel now and you will see an event sent successfully.

refund

I added some notes to the readme, it confused me also at first. https://github.com/PrestaShop/ps_googleanalytics/blob/dev/Readme.md

SimonGrn:
I'm not able to make this works :cry: 
Here are my steps:
- `git clone` PrestaShop
- switch on the 1.7.7.x branch
- install it
- go to `/modules`
- `git clone` the ps_googleanalytics module
- go into the directory, `git prc` to switch to this PR (43) on the `dev` branch
- install the module in the BO (version is 3.3.0 
image
and MySQL tables are correct
image
)
- set up the module with my own GA analytics code created only for this purpose, for the correct shop URL (my shop is installed on my own VPS, like a production environment)
- create an order from scracth, status on "Awaiting check payment"
- on the order page, start recording with GTag Assistant 
- change the status to "canceled"
image
- stop recording (GTag Assistant is showing me an error at this point)

This is the result in the GTag Assistant modal:
image

This is the only info I have, no "Event" anywhere.
On my Google Analytics console, I can see myself browsing the shop on the RealTime tab:
image
(I know it's me since I can see the URL I'm browsing). So I know the Google Tag is working properly.

Here is the full report from the GTag Assistant:
image

The table is still empty:
image


Do you see something I'm missing ?
Thanks !

Hlavtox:
@SimonGrn Try it in 1.7.6.5 please. 

In 1.7.7, it seems that hookdisplayAdminOrder is not working at all.

SimonGrn:
**TESTING IN 1.7.6.5**
After changing the status to "Canceled" in the BO, I have the correct event captured:
image

The MySQL data is correct (before and after the status change):
image

After changing the status to "Awaiting check payment" then to "Canceled" again, there is no new Event fired:
image
And database is still okay
image


This is correct for 1.7.6.5.
We still have the problem of this PR not working in 1.7.7. I'm gonna check with a developer to see if it's an issue with 1.7.7 or a deeper problem.
Thanks !

SimonGrn:
Regarding the [scope of compatibility](https://github.com/PrestaShop/ps_googleanalytics/blob/master/ps_googleanalytics.php#L49) for this module, I can't :heavy_check_mark: this Pr if it's not working on the next version of PrestaShop which will drop soon.
I suggest you sort the hook problem in your PR (it's apparently because 1.7.7 introduced more strict hook system that doesn't allow `echo` and will work fine with `return` ?) then update this PR !
We'll be happy to review it for 1.7.7 version and :heavy_check_mark: it :)

Thanks for your comprehension.

Hlavtox:
@SimonGrn Simon, thanks, but I will need your assistance.

The echo problem is solved easily by changing it to:
```
$output = $this->_runJs($this->context->cookie->ga_admin_refund, 1);
unset($this->context->cookie->ga_admin_refund);
return $output;
```

But there is some problem with **$this->context->cookie->ga_admin_refund** variable, which was already present in the module.

This is the simplified code:
```
    public function hookActionOrderStatusPostUpdate(array $params)
    {
        $this->context->cookie->ga_admin_refund = "AAAAA";
        return;
    }

    public function hookdisplayAdminOrder()
    {
        return $this->context->cookie->ga_admin_refund;
    }

```

I go to order screen, I change the order status and it displays "TEST", a completely different string I somehow got into it?

SimonGrn:
Oh my this is a little bit too technical for me, I suggest you create an issue or something because I have absolutely no idea of how this works I'm sorry :sweat_smile:

Hlavtox:
Ping @PierreRambaud, can you have a look above? https://github.com/PrestaShop/ps_googleanalytics/pull/43#issuecomment-650234867

PierreRambaud:
> Ping @PierreRambaud, can you have a look above? [#43 (comment)](https://github.com/PrestaShop/ps_googleanalytics/pull/43#issuecomment-650234867)

Unfortunately, I can't have a look before one week :sweat_smile: (Leaving GitHub for my mental wellness #vacations)

apacios:
A new version of this module is waiting for QA : https://github.com/PrestaShop/ps_googleanalytics/pull/50

In this version I modified the whole structure of the module so that it can be more easily to maintain. Also, the module comply with the PSR-2 standard.

I think that if it's necessary to make modifications on this PR the best would be to integrate it starting from this version to come.

Hlavtox:
@apacios Ok, should I wait until your PR is merged?

PS: Are you sure this will work? You kept echo in the hook. https://github.com/PrestaShop/ps_googleanalytics/blob/07563209a788a151730fc2ff5f3a678d7af47bb9/ps_googleanalytics.php#L142

apacios:
@Hlavtox I tested the whole module and all seems to work fine. But as it's waiting for QA I think you should wait a bit.
The QA will test my PR in the upcoming days

matks:
>         return $this->context->cookie->ga_admin_refund;

Hi @Hlavtox I finally got some time to look at this.

So indeed as you have seen we cannot echo anymore in migrated page, a `return` statement is needed.

If I am not wrong the remaining issue is this one:
```
public function hookActionOrderStatusPostUpdate(array $params)
    {
        $this->context->cookie->ga_admin_refund = "AAAAA";
        return;
    }

    public function hookdisplayAdminOrder()
    {
        return $this->context->cookie->ga_admin_refund;
    }
```

You expect `hookdisplayAdminOrder` to print the content of the variable `$this->context->cookie->ga_admin_refund`. But this is not working.

And I think it's expected 😄 . Because **when modifying the order status, a POST HTTP request is sent, handled by PrestaShop backend, then PrestaShop backend sends a HTTP 302 response to browser, browser sends a HTTP GET request and obtains a refreshed version of the page.**

So on PS 1.7.7 your 2 hooks are not called within the same HTTP requests, and since PHP is being killed between the 2 requests the variable is lost.

Hlavtox:
@matks Ooooh, so that's why. I thought that saving something to prestashop context->cookie would automatically handle saving the cookie itself.

And if you confirm the echo not working, then @apacios 's https://github.com/PrestaShop/ps_googleanalytics/blob/07563209a788a151730fc2ff5f3a678d7af47bb9/ps_googleanalytics.php#L142 should not work, hm? ❓

PR #44, opened 2020-03-04
Update Readme.md

Hlavtox:
I added a section to readme to inform about what seems to be an error. Several issues have been opened in the past, including one of mine.


PierreRambaud:
Thanks @Hlavtox

PR #45, opened 2020-03-09
 Escape some SQL query

Feloot:
Escape the $data var in _manageData() using pSQL()


PierreRambaud:
No need QA, thanks @Feloot

PR #66, opened 2020-08-26
Add tracking of cancelled orders

Hlavtox:
This PR enables tracking of cancelled orders, which could greatly affect the statistics. The refund functionality is already provided by the enhanced ecommerce library, but never used.

I added a configuration field to module settings, where the user can select in which order states, he wants the order to be refunded (cancelled) in Google Analytics.

When the order status changes to one of these, it sends the information to google analytics.

**Technical changes**
- The functionality hooks to actionOrderStatusPostUpdate.
- Adds a configuration field and an entry in database with encoded order states.
- Adds 1 column to ps_ganalytics to prevent multiple refunds for one order. 
- Bumps module version to 4.1.0 to run the upgrade script needed - to upgrade the database and insert a configuration field.
- Fixes forgotten echo and cookie->write in hookDisplayAdminOrder.

**How it works**
- After you change the order status, the module checks, if the new status is one of the states the user wants the order to be cancelled.
- If so, it checks the database, if this order was not cancelled before - to prevent multiple refunds, which Google Analytics unfortunately allows.
- If no refund information found in database, adds the refund action by MBG.refundByOrderId.

**How to test**
1. Install Google Tag Assistant (in Chrome).
2. Install ps_googleanalytics module, set it up. Add some states to the configuration field.
3. Make an order, go to it's administration.
4. Enable recording in Google Tag Assistant and refresh page.
5. Change order status to cancelled.
6. After page refresh, check Google Tag Assistant and see the Refund event being sent.
7. Change order status to "Processing", then, change it back to "Cancelled".
8. No refund event should be sent.

The refund will appear after minutes/hours in "Refund" column in Google Analytics.


jf-viguier:
You rocks @Hlavtox thanks

classes/Hook/HookActionOrderStatusPostUpdate.php
```diff
@@ -0,0 +1,82 @@
...
+     *
+     * @return void
+     */
+    public function run()
+    {
+        // If we have the order AND the new order status being set AND the order status is changing to CANCELLED
+        if (empty($this->params['id_order'])
+        || empty($this->params['newOrderStatus']->id)
+        || (int) $this->params['newOrderStatus']->id !== (int) Configuration::get('PS_OS_CANCELED')
+        ) {
```
PierreRambaud, line 51:
```suggestion
        if (empty($this->params['id_order'])
            || empty($this->params['newOrderStatus']->id)
            || (int) $this->params['newOrderStatus']->id !== (int) Configuration::get('PS_OS_CANCELED')
        ) {
```

classes/Hook/HookActionOrderStatusPostUpdate.php
```diff
@@ -0,0 +1,82 @@
...
+
+        // We check if the refund was already sent to Google Analytics
+        $gaRefundSent = Db::getInstance()->getValue(
+            'SELECT id_order FROM `' . _DB_PREFIX_ . 'ganalytics` WHERE id_order = ' . (int) $this->params['id_order'] . ' AND refund_sent = 1'
+        );
+
+        // If it was not already refunded
+        if ($gaRefundSent === false) {
+            // We refund it and set the "sent" flag to true
+            $this->context->cookie->__set('ga_admin_refund', 'MBG.refundByOrderId(' . json_encode(['id' => $this->params['id_order']]) . ');');
```
PierreRambaud, line 63:
```suggestion
            $this->context->cookie->ga_admin_refund = 'MBG.refundByOrderId(' . json_encode(['id' => $this->params['id_order']]) . ');';
```

upgrade/upgrade-4.1.0.php
```diff
@@ -0,0 +1,37 @@
...
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+if (!defined('_PS_VERSION_')) {
+    exit;
+}
+
+function upgrade_module_4_1_0($object)
+{
+    if ($object->registerHook('actionOrderStatusPostUpdate') == false) {
```
PierreRambaud, line 32:
```suggestion
    if (!$object->registerHook('actionOrderStatusPostUpdate')) {
```

ps_googleanalytics.php
```diff
@@ -139,11 +139,14 @@ public function hookDisplayAdminOrder()
...
 
-        echo $gaTagHandler->generate(
+        $output = '';
+        $output = $gaTagHandler->generate(
             $this->context->cookie->__get('ga_admin_refund'),
             true
         );
         $this->context->cookie->__unset('ga_admin_refund');
+
+        return $output;
```
PierreRambaud, line 149:
```suggestion
        return $gaTagHandler->generate(
            $this->context->cookie->__get('ga_admin_refund'),
            true
        );
        $this->context->cookie->__unset('ga_admin_refund');
```

But you should be sure it needs a return and not an echo :sweat_smile:

Hlavtox: Yes, echo does not work on migrated pages.
Maybe I am wrong, but with your suggestion, how will the cookie be unset, when you return before?
The code would then be run infinite times, hm?

classes/Hook/HookActionOrderStatusPostUpdate.php
Hlavtox: @PierreRambaud This is how it's done in the original code https://github.com/PrestaShop/ps_googleanalytics/blob/master/classes/Hook/HookActionProductCancel.php

PierreRambaud: Sad, when it's a magic function :(

ps_googleanalytics.php
PierreRambaud: Ahah you're right, didn't see the cookie unset, badly scrolling :joy:

PierreRambaud: About the echo, you should have a test on PS 1.6 at least to be sure :)

classes/Form/ConfigurationForm.php
```diff
@@ -159,6 +174,11 @@ public function generate()
         $helper->fields_value['GA_CROSSDOMAIN_ENABLED'] = Configuration::get('GA_CROSSDOMAIN_ENABLED');
         $helper->fields_value['GA_ANONYMIZE_ENABLED'] = Configuration::get('GA_ANONYMIZE_ENABLED');
 
+        $gaCancelledStates = json_decode(Configuration::get('GA_CANCELLED_STATES'));
+        if (!empty($gaCancelledStates) || $gaCancelledStates !== null) {
```
PierreRambaud, line 178:
```suggestion
        if (!empty($gaCancelledStates)) {
```

classes/Form/ConfigurationForm.php
```diff
@@ -159,6 +174,11 @@ public function generate()
         $helper->fields_value['GA_CROSSDOMAIN_ENABLED'] = Configuration::get('GA_CROSSDOMAIN_ENABLED');
         $helper->fields_value['GA_ANONYMIZE_ENABLED'] = Configuration::get('GA_ANONYMIZE_ENABLED');
 
+        $gaCancelledStates = json_decode(Configuration::get('GA_CANCELLED_STATES'));
```
PierreRambaud, line 177: json_encode returns an StdClass instance, are you sure? Shouldn't be an array?

classes/Form/ConfigurationForm.php
```diff
@@ -196,6 +217,15 @@ public function treat()
             $treatmentResult .= $this->module->displayConfirmation($this->module->l('Settings for Anonymize IP updated successfully'));
         }
 
+        if (null !== $gaCancelledStates) {
```
PierreRambaud, line 220: I'm not sure the value can be null because of `Tools::getValue` :thinking:

classes/Hook/HookActionOrderStatusPostUpdate.php
```diff
@@ -0,0 +1,88 @@
...
+namespace PrestaShop\Module\Ps_Googleanalytics\Hooks;
+
+use Configuration;
+use Context;
+use Db;
+use Ps_Googleanalytics;
+
+class HookActionOrderStatusPostUpdate implements HookInterface
+{
+    private $module;
```
PierreRambaud, line 30:
```suggestion
    /**
     * @var Ps_Googleanalytics
     */
    private $module;
```

classes/Hook/HookActionOrderStatusPostUpdate.php
```diff
@@ -0,0 +1,88 @@
...
+    private $context;
```
PierreRambaud, line 31:
```suggestion
    /**
     * @var Context
     */
    private $context;
```

classes/Hook/HookActionOrderStatusPostUpdate.php
```diff
@@ -0,0 +1,88 @@
...
+    private $params;
```
PierreRambaud, line 32:
```suggestion
    /**
     * @var array
     */
    private $params;
```

classes/Hook/HookActionOrderStatusPostUpdate.php
```diff
@@ -0,0 +1,88 @@
...
+     */
+    public function run()
+    {
+        // If we do not have an order or a new order status, we return
+        if (empty($this->params['id_order']) || empty($this->params['newOrderStatus']->id)) {
+            return;
+        }
+
+        // We get all states in which the merchant want to have refund sent and check if the new state being set belongs there
+        $gaCancelledStates = json_decode(Configuration::get('GA_CANCELLED_STATES'));
```
PierreRambaud, line 53:
```suggestion
        $gaCancelledStates = json_decode(Configuration::get('GA_CANCELLED_STATES'));
```
same as previous, it's not an array, it's an object

classes/Hook/HookActionOrderStatusPostUpdate.php
```diff
@@ -0,0 +1,88 @@
...
+        if (empty($gaCancelledStates) ||
+            $gaCancelledStates == null ||
+            !in_array($this->params['newOrderStatus']->id, $gaCancelledStates)
```
PierreRambaud, line 56:
```suggestion
        if (empty($gaCancelledStates) ||
            !in_array($this->params['newOrderStatus']->id, $gaCancelledStates)
```

classes/Form/ConfigurationForm.php
Hlavtox: @PierreRambaud If the value does not exist in database, it returns NULL, so, maybe better be safe?

classes/Hook/HookActionOrderStatusPostUpdate.php
Hlavtox: Fixed

classes/Hook/HookActionOrderStatusPostUpdate.php
Hlavtox: Fixed

classes/Hook/HookActionOrderStatusPostUpdate.php
Hlavtox: Fixed

classes/Hook/HookActionOrderStatusPostUpdate.php
Hlavtox: Fixed

classes/Form/ConfigurationForm.php
Hlavtox: Fixed

classes/Hook/HookActionOrderStatusPostUpdate.php
Hlavtox: Same as the other one, what if the value is deleted from the database?

classes/Form/ConfigurationForm.php
Quetzacoalt91:
```
<?php

$gaCancelledStates = null;

$isEmpty = empty($gaCancelledStates);

var_dump($isEmpty);
```

Result:
```
bool(true)
```

The second check is redundant with the first one and can be removed.

Hlavtox: OK, done guys, thanks!

Hlavtox:
@PierreRambaud @Quetzacoalt91 All suggestions implemented guys 👍

classes/Hook/HookActionOrderStatusPostUpdate.php
```diff
@@ -0,0 +1,96 @@
...
+            );
+        }
+    }
+
+    /**
+     * setParams
+     *
+     * @param array $params
+     */
+    public function setParams($params)
```
PierreRambaud, line 92:
```suggestion
    public function setParams(array $params)
```

PierreRambaud: Can you test with this change?

Hlavtox: @PierreRambaud All other hooks are written as I did, shouldn't we keep it consistent and maybe change it in another PR?

PierreRambaud: It's better in a new PR :+1:

PierreRambaud (approved):

classes/Hook/HookActionOrderStatusPostUpdate.php
```diff
@@ -0,0 +1,96 @@
...
+        if (empty($gaCancelledStates) || !in_array($this->params['newOrderStatus']->id, $gaCancelledStates)) {
+            return;
+        }
+
+        // We check if the refund was already sent to Google Analytics
+        $gaRefundSent = Db::getInstance()->getValue(
+            'SELECT id_order FROM `' . _DB_PREFIX_ . 'ganalytics` WHERE id_order = ' . (int) $this->params['id_order'] . ' AND refund_sent = 1'
+        );
+
+        // If it was not already refunded
```
matks, line 74: Very nice comments, this makes the code very easy to read 👍

matks (approved):
PR looks good to me :)

Since it is quite a big change I think we'll need QA to validate it.

classes/Hook/HookActionOrderStatusPostUpdate.php
Hlavtox: @matks Thanks! Trying to keep WTFs of somebody who reads it to minimum 😄

sarahdib:
Hello @Hlavtox 

One last change and it's ok :) 
If the cancel status isn't configure in ps_googleanalytics there is a warning 
Capture d’écran 2020-09-17 à 16 50 40
I think this can be display with a message above the field instead of a warning

Hlavtox:
@sarahdib Thanks! I will resolve the issue and let you know. :-)

The PR won't work in current dev anyway until https://github.com/PrestaShop/ps_googleanalytics/pull/77 is merged.

upgrade/upgrade-4.1.0.php
```diff
@@ -0,0 +1,43 @@
...
+{
+    if (!$object->registerHook('actionOrderStatusPostUpdate')) {
+        return false;
+    }
+
+    $defaultGaCancelledStates = [
+      0 => Configuration::get('PS_OS_CANCELED'),
+    ];
+
+    Configuration::updateValue('GA_CANCELLED_STATES', json_encode($defaultGaCancelledStates));
```
kpodemski, line 40: @Hlavtox 

you can do:
`Configuration::updateValue('GA_CANCELLED_STATES', json_encode([Configuration::get('PS_OS_CANCELED')]));`

Hlavtox: Thanks! ;-)

Hlavtox:
@sarahdib Ready to QA! :-)

Hlavtox:
@sarahdib @SD1982 @matks Conflicts resolved after last PR's. You can test it. I hope it will be OK. 🙏 🚀

PierreRambaud (approved):

Progi1984:
Thanks @Hlavtox & @sarahdib

Issue #61, opened 2020-08-17
GA events are fired twice in BackOffice


HartLarsson:
THe GA JS code is persent also in the backoffice.

PS 1.7.6.7
GA 4.0.0

GA JS code must be fired only in frontoffice.

matks:
This is not a bug, it's a feature.

See the README:

> Google Tag Assistant will report "No HTTP Response detected" error in Prestashop back-office. This is not a bug. This happens, because the module does not send page views in BO, not to influence statistics. Only events, like refunds, are sent from BO.

HartLarsson:
strange feature because in Analytics i can see my backoffice pages!
![image](https://user-images.githubusercontent.com/16102074/90383543-947ee000-e080-11ea-943d-d902ea86b285.png)

I hope this feature, that for me is a bug (tracking BO pages) must be removed, also because is fired TWICE!

matks:
> strange feature because in Analytics i can see my backoffice pages!

There might be weird behaviors, bugs or issues in how the feature is implemented :)

> I hope this feature, that for me is a bug (tracking BO pages) must be removed

No it will not be removed :) . Some people use this feature. If you do not like it you can modify the module to disable it in the back-office.

This module is free. It should provide as much as possible features for users, so that everybody can use what they want. Of course if you dont want to use some of its features, since it's open-source you can customize it to fit your needs 😄

HartLarsson:
"No it will not be removed :) . Some people use this feature. If you do not like it you can modify the module to disable it in the back-office."
an you lease tell me how to remove the BO tracking or to fix the tracking of all BO pages?

HartLarsson:
> There might be weird behaviors, bugs or issues in how the feature is implemented :)

for sure there is something that not work, because is ok tracking refunds but not all pages :)

HartLarsson:
seems that in the ps_googleanalytics.tpl this part of the code is always FALSE

`    {if $backOffice}
        ga('set', 'nonInteraction', true);
    {else}
        ga('send', 'pageview');
    {/if}`

is why a pageview is always fired in BO while in previous 3,2.0 was working correctly

Hlavtox:
@matks it's a bug, it cannot track BO pages, it does so after refactoring

matks:
@HartLarsson I have modified this issue title to "GA events are fired twice" as this is the real bug for me.

If you think it makes sense, you can open an issue "We should be able to disable Back-Office tracking".

[...]

Issue #64, opened 2020-08-26
Pageviews are sent from backoffice (3 times!!!)


Hlavtox:
New version 4.0.0 is sending pageviews from the backoffice.
What's more wrong, it does _ga('send', 'pageview');_ 3 times.

This influences statistics and should be fixed asap.

matks:
@Hlavtox I have checked the behavior. I can acknowledge the GA event "pageview" is fired twice on BO pages because of a bug.

The PR will fix this https://github.com/PrestaShop/ps_googleanalytics/pull/67

However I do not see 3 events being fired. Can you provide a detailed scenario of how you have setup your shop and module to see that ? I cant see how you could get 3 events.

Hlavtox:
@matks Open BO order view page in 1.7.6.

The bug is present in 1.7.7 also, but you must fix echo to return in hookDisplayAdminOrder in the module.

Hlavtox:
![Výstřižek](https://user-images.githubusercontent.com/6097524/91446618-f22cdc80-e877-11ea-878d-3a9f4113e1b5.PNG)

Hlavtox:
![1](https://user-images.githubusercontent.com/6097524/91448605-5f417180-e87a-11ea-83e9-72d611c0ef5a.PNG)

Hlavtox:
Fixed by https://github.com/PrestaShop/ps_googleanalytics/pull/77

matks:
Fix finally available in new module version https://github.com/PrestaShop/ps_googleanalytics/releases/tag/v4.1.0

PR #67, opened 2020-08-27
Fix logic, avoid double pageview for BO pages

matks:
| Questions     | Answers
| Description?  | We got reports that the v4.0.0 of the module does send double pageviews to GA. I checked the code and found that, because of a bad variable assignment, the context of HookDisplayHeade was not correctly set. I fixed that, which should resume expected behavior for the module = pageview event only sent once. See changed files for more detail.
| Type?         | bug fix
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/ps_googleanalytics/issues/61 and probably https://github.com/PrestaShop/ps_googleanalytics/issues/64
| How to test?  | See below

### See error

1. Install module, set it up with Google Analytics account.
2. Open a BO page, and you can see the number of pageviews is x2 . If you open the page 3 times you can see 6 pageviews.

### See fix

3. Test the module with this PR. The pageviews number is correct.

### DIFF
--- a/classes/Hook/HookDisplayHeader.php
+++ b/classes/Hook/HookDisplayHeader.php
@@ -31,6 +31,10 @@ class HookDisplayHeader implements HookInterface
     private $module;
     private $context;
     private $params;
+
+    /**
+     * @var bool
+     */
     private $backOffice;
 
     public function __construct(Ps_Googleanalytics $module, Context $context)
@@ -100,12 +104,18 @@ class HookDisplayHeader implements HookInterface
     }
 
     /**
-     * setBackOffice
-     *
-     * @param array $backOffice
+     * @param bool $backOffice
      */
     public function setBackOffice($backOffice)
     {
-        $this->module->backOffice = $backOffice;
+        $this->acknowledgeBackOfficeContext($backOffice);
+    }
+
+    /**
+     * @param bool $isBackOffice
+     */
+    public function acknowledgeBackOfficeContext($isBackOffice)
+    {
+        $this->backOffice = $isBackOffice;
     }
 }


classes/Hook/HookDisplayHeader.php
```diff
@@ -100,12 +104,18 @@ public function setParams($params)
...
 
     /**
-     * setBackOffice
-     *
-     * @param array $backOffice
+     * @param bool $backOffice
      */
     public function setBackOffice($backOffice)
     {
-        $this->module->backOffice = $backOffice;
```
matks, line 109: So the right variable to assign is `$this->backOffice`

classes/Hook/HookDisplayHeader.php
```diff
@@ -100,12 +104,18 @@ public function setParams($params)
...
+        $this->acknowledgeBackOfficeContext($backOffice);
+    }
+
+    /**
+     * @param bool $isBackOffice
+     */
+    public function acknowledgeBackOfficeContext($isBackOffice)
```
matks, line 117: I think function name `setBackOffice` is ambiguous, which is why I provide a replacement. I keep the old function for backward compatibility.

atomiix (approved):

PierreRambaud (approved):

Hlavtox:
@matks @PierreRambaud You need to fix the third one also.

hookDisplayAdminOrder -> $gaTagHandler->generate outputs another pageview in ga_tag.tpl

Progi1984:
Thanks @matks & @sarahdib

Issue #62, opened 2020-08-19
Bad date_add


Daaaaad:
Hello,

I just installed and configured the module on a PS 1.6.1.24.

When I pass an order the data registered in ps_ganalytics is the following : 

<img width="993" alt="Capture d’écran 2020-08-19 à 11 36 06" src="https://user-images.githubusercontent.com/167037/90618692-a2636b00-e210-11ea-86d3-eedfd94b52d3.png">

"date_add" contain a bad value... like if the NOW() function was not interpreted.

### How to reproduce

1. Install ps_googleanalytics v4.0.0
2. Set it up
3. Create an order from FO
4. Open a SQL client (phpMyAdmin, Sequel Pro ...) and look at content of SQL table `ps_ganalytics`
5. See content of column `date_add` is 0000-00-00 00:00:00

Daaaaad:
The SQL query executed to add those rows is : 

INSERT IGNORE INTO `ps_ganalytics` (`id_order`, `id_shop`, `sent`, `date_add`) VALUES ('4405', '1', '0', 'NOW()')

In this case NOW() is not interpreted correctly 😞

matks:
I confirm this issue.

matks:
I submitted a PR to fix this bug https://github.com/PrestaShop/ps_googleanalytics/pull/68

Daaaaad:
Hello @matks, will you fix it in other places?

matks:
> Hello @matks, will you fix it in other places?

I guess I have no choice, right 😛  ?

Shoprunners:
Hi, without bugfixing the other places, this module will not work since there is no order data in the database.

matks:
Fix finally available in new module version https://github.com/PrestaShop/ps_googleanalytics/releases/tag/v4.1.0

PR #68, opened 2020-08-27
Handle NOW correctly when creating order and register in SQL

matks:
| Questions     | Answers
| Description?  | When placing an order, the timestamp data was badly registered in MySQL (0000-00-00 00:00:00 instead of current datetime) inside `ps_ganalytics` table. It's because `NOW()` is specific value and must be configured as such, else PrestaShop considers it's the string `NOW()` and not the operator.
| Type?         | bug fix
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/ps_googleanalytics/issues/62
| How to test?  | See issue

### DIFF
--- a/controllers/front/ajax.php
+++ b/controllers/front/ajax.php
@@ -41,7 +41,7 @@ class ps_GoogleanalyticsAjaxModuleFrontController extends ModuleFrontController
         (new GanalyticsRepository())->updateData(
             [
                 'sent' => 1,
-                'date_add' => 'NOW()',
+                'date_add' => ['value' => 'NOW()', 'type' => 'sql'],
             ],
             'id_order = ' . $orderId,
             1


atomiix (approved):

PierreRambaud (approved):

Daaaaad:
Hi @matks, I think there are other places in the module where the NOW() function is not interpreted correctly (In the HookDisplayBackOfficeHeader class, in the HookDisplayOrderConfirmation class, etc)

matks:
> Hi @matks, I think there are other places in the module where the NOW() function is not interpreted correctly (In the HookDisplayBackOfficeHeader class, in the HookDisplayOrderConfirmation class, etc)

Indeed. 😞

Progi1984:
Thanks @sarahdib & @matks

Issue #71, opened 2020-08-27
Feature request: allow to disable Back Office pages tracking


matks:
We got reports that tracking BO pages is not suitable for all merchants so I suggest adding a simple option to disable Back Office pages tracking.

matks:
Fixed by https://github.com/PrestaShop/ps_googleanalytics/pull/72

matks:
Fix finally available in new module version https://github.com/PrestaShop/ps_googleanalytics/releases/tag/v4.1.0

PR #72, opened 2020-08-27
Add Configuration option to disable Back Office tracking

matks:
| Questions     | Answers
| Description?  | Add an option inside Configure page to disable Back Office tracking
| Type?         | new feature
| BC breaks | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/ps_googleanalytics/issues/71
| How to test?  | Open module, see the new option. Please check that when used (= disable BO tracking), no data is sent to Google from BO pages.

### DIFF
--- a/classes/Form/ConfigurationForm.php
+++ b/classes/Form/ConfigurationForm.php
@@ -127,6 +127,24 @@ class ConfigurationForm
                         ],
                     ],
                 ],
+                [
+                    'type' => 'switch',
+                    'label' => $this->module->l('Disable Back Office Tracking'),
+                    'name' => 'GA_TRACK_BACKOFFICE_DISABLED',
+                    'hint' => $this->module->l('Use this option to disable the tracking inside the Back Office'),
+                    'values' => [
+                        [
+                            'id' => 'ga_track_backoffice',
+                            'value' => 1,
+                            'label' => $this->module->l('Enabled'),
+                        ],
+                        [
+                            'id' => 'ga_do_not_track_backoffice',
+                            'value' => 0,
+                            'label' => $this->module->l('Disabled'),
+                        ],
+                    ],
+                ],
             ],
             'submit' => [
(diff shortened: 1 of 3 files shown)


PierreRambaud (approved):

kpodemski:
I'm not sure if that's a good fix? FWIK people want to stop sending page view, right? this PR will remove module from working in BO completely, which means that handling cancelled order etc. will not work, right?

matks:
> I'm not sure if that's a good fix? FWIK people want to stop sending page view, right? this PR will remove module from working in BO completely, which means that handling cancelled order etc. will not work, right?

I think both needs exist.

[Here you have](https://github.com/PrestaShop/ps_googleanalytics/issues/61#issue-680060310) someone who believes the module should only track users in FO.
And [here you have](https://github.com/PrestaShop/ps_googleanalytics/issues/64) someone who believes the module should not send pageview from the FO.

Both make sense. Actually, depending on how people use GA, different configurations could become relevant. If you use GA only to track users journey and sources, you dont need it in BO. If you use GA to monitor your ecommerce KPIs, then you need indeed the "cancel order" events.

I think this module is currently doing things in a way quite ... brutal 😛 although it could, provide deep and accurate customization/configuration abilities.

I dont know however if I'll have the needed time to address all that. Today I spent multiple unscheduled hours on this module because I saw the issues piling up 😅 but I am now late on my other assignments.

matks:
Today I wanted to address the emergencies only. However a bigger thinking is required. And bigger thinking can only happen when all knowledge is provided. So I need https://github.com/PrestaShop/ps_googleanalytics/issues/74 is a 1st necessary step.

Hlavtox:
@matks You cannot kill the module like this.

The JS code is needed to provide refund functions. 

**Only the pageviews must be disabled.**

Progi1984:
Thanks @matks & @sarahdib

PR #75, opened 2020-08-31
Fix double and triple pageview requests

Hlavtox:
| Questions     | Answers
| Description?  | There is an useless code that only outputs extra pageviews to the page. Pageview is always sent from header hook (ps_googleanalytics.tpl).
| Type?         | bug fix
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/ps_googleanalytics/issues/64 https://github.com/PrestaShop/ps_googleanalytics/issues/61
| How to test?  | Install Google Tag Assistant. Open any order in BO, open any page in FO. Enable assistant. Refresh page. Only 1 pageview will be sent.

### DIFF
--- a/views/templates/hook/ga_tag.tpl
+++ b/views/templates/hook/ga_tag.tpl
@@ -28,11 +28,3 @@
     </script>
     {/literal}
 {/if}
-
-{if ($jsState != 1 && $isBackoffice === true)}
-    {literal}
-    <script type="text/javascript">
-        ga('send', 'pageview');
-    </script>
-    {/literal}
-{/if}


PierreRambaud (approved):

matks (changes requested):
I'm sorry I reject this PR.

Unless proven otherwise*, this part of the code is not useless.

In `ps_googleanalytics.tpl` the pageView is supposed to be sent only on front-office
```
{if $backOffice}
        ga('set', 'nonInteraction', true);
    {else}
        ga('send', 'pageview');
    {/if}
```

You cannot just remove some code because you dont like it (well, actually _you_ can but I will not approve the PR which does this).

Since there is a file HookDisplayBackOfficeHeader.php we can see the module is _built_ to be active in the Back-Office. You cant just remove it like this 😄  . As we discussed before, we can provide a "feature flag" option to disable this, we can explore how to fine-tune this ... but let's not just throw it away.

<hr>

*Note: if you can prove, factually, this code is 100% useless I will agree to remove it. But so far I have not seen such proof.

Hlavtox:
@matks The code is useless and it will continue to work properly, for both BO and FO. Let me explain why.

If you apply https://github.com/PrestaShop/ps_googleanalytics/pull/72 with my proposed change _(otherwise you kill GA refunds in BO)_ AND you fix **backOffice** variable with your other PR https://github.com/PrestaShop/ps_googleanalytics/pull/67, it will work.

The pageview request will be provided from ps_googleanalytics.tpl. Both in FO and BO. If you read me proposal, this would work properly:

```
{if $backOffice && $trackBackOfficeDisabled}
    ga('set', 'nonInteraction', true);
  {else}
    ga('send', 'pageview');
{/if}
```

Issue #74, opened 2020-08-27
Documentation: provide a documentation about how this module work and what data is sent to GA


matks:
In order to be able to provide accurate and extended configuration capabilities / data synchronization between PS and GA, it would be useful to provide a documentation about how this module work and what data is sent to GA.

The doc should explain the inner working of the module.
The doc should explain what data are being sent from PS to GA, and where.

Once the list is complete, it'll provide an accurate overview of what data is missing / can be expanded (see https://github.com/PrestaShop/ps_googleanalytics/issues/57) and maybe open the path for advanced configuration settings (being able, from BO, to enable/disable what is being sent)

Hlavtox:
In my opinion, the data the module should send is clearly defined in Google Analytics's documentation and the module should respect that.

**For example, the discussion regarding taxes.**
https://developers.google.com/analytics/devguides/collection/analyticsjs/enhanced-ecommerce#measuring-transactions

```
ga('ec:addProduct', {   
  'id': 'P12345',      
  'name': 'Android Warhol T-Shirt',
  'category': 'Apparel',     
  'brand': 'Google',        
  'variant': 'black',        
  'price': '29.20',  // Product price without tax
  'coupon': 'APPARELSALE',      
  'quantity': 1      
});

ga('ec:setAction', 'purchase', {     
  'id': 'T12345',  
  'affiliation': 'Google Store - Online', 
  'revenue': '37.39',  // Order total with tax
  'tax': '2.85',  // Order total tax
  'shipping': '5.34',  // Order total shipping with tax
  'coupon': 'SUMMER2013' 
});
```

Kharny:
I agree with this issues. Actually I do no how to install the module to start. I did not find the information except configuration: 1 install the module.

PR #77, opened 2020-09-17
Fix backoffice JS code, fix doubled and tripled pageviews, fix echo in hook, fix default BO tracking - for versions up to 1.7.6

Hlavtox:
# This PR fixes following bugs and confusions

### JS code that was killed before page output
- JS code that is not included in a backoffice header because of https://github.com/PrestaShop/ps_googleanalytics/pull/72/. This PR changes the logic and disables only the pageview itself in the TPL.
- This logic needs to work correctly because other functionality depends on it.
-- Default actionProductCancel functionality of the module
-- Order refund functionality I am adding here: https://github.com/PrestaShop/ps_googleanalytics/pull/66

### Backoffice views enabled by default
- Until 4.0.0, this module did not track backoffice views.
- There was a bug in 4.0.0 that enabled backoffice pageviews to be sent.
- @matks decided to keep the tracking there and disable it when wanted.
- Based on previous behavior, the default state should be that the module should not track backoffice views. 
- If users upgrade to version 4, they would not be happy that it's behaving differently without their knowledge and they have to go to module settings to disable it.

### Doubled and tripled pageviews
- Removes a code that sends two or three extra pageviews in BO. This code is not present in v3. Pageview is sent in the header, no need to have it in the other TPL. This code was not present in v3 of the module.

### Code output to order admin page was not working
- Fixes an echo in one hook. 
- Echo does not work on migrated pages and needs to be changed to return.

### How to test
- Install Google Tag Assistant
- Try tracking in FO - _properly included, 1 pageview_
- Try tracking in BO with BO tracking enabled - _properly included, 1 pageview will be sent_
- Try tracking in BO with BO tracking disabled - _properly included, no pageview is sent (reports no HTTP response, that's normal)_
- Try triggering hookActionProductCancel, refund code will be output on BO order view page, if needed.

### DIFF
--- a/classes/Form/ConfigurationForm.php
+++ b/classes/Form/ConfigurationForm.php
@@ -129,9 +129,9 @@ class ConfigurationForm
                 ],
                 [
                     'type' => 'switch',
-                    'label' => $this->module->l('Disable Back Office Tracking'),
-                    'name' => 'GA_TRACK_BACKOFFICE_DISABLED',
-                    'hint' => $this->module->l('Use this option to disable the tracking inside the Back Office'),
+                    'label' => $this->module->l('Enable Back Office Tracking'),
+                    'name' => 'GA_TRACK_BACKOFFICE_ENABLED',
+                    'hint' => $this->module->l('Use this option to enable the tracking inside the Back Office'),
                     'values' => [
                         [
                             'id' => 'ga_track_backoffice',
@@ -176,7 +176,7 @@ class ConfigurationForm
         $helper->fields_value['GA_USERID_ENABLED'] = Configuration::get('GA_USERID_ENABLED');
         $helper->fields_value['GA_CROSSDOMAIN_ENABLED'] = Configuration::get('GA_CROSSDOMAIN_ENABLED');
         $helper->fields_value['GA_ANONYMIZE_ENABLED'] = Configuration::get('GA_ANONYMIZE_ENABLED');
-        $helper->fields_value['GA_TRACK_BACKOFFICE_DISABLED'] = Configuration::get('GA_TRACK_BACKOFFICE_DISABLED');
+        $helper->fields_value['GA_TRACK_BACKOFFICE_ENABLED'] = Configuration::get('GA_TRACK_BACKOFFICE_ENABLED');
 
         return $helper->generateForm($fields_form);
     }
@@ -193,7 +193,7 @@ class ConfigurationForm
         $gaUserIdEnabled = Tools::getValue('GA_USERID_ENABLED');
         $gaCrossdomainEnabled = Tools::getValue('GA_CROSSDOMAIN_ENABLED');
         $gaAnonymizeEnabled = Tools::getValue('GA_ANONYMIZE_ENABLED');
-        $gaTrackBackOffice = Tools::getValue('GA_TRACK_BACKOFFICE_DISABLED');
+        $gaTrackBackOffice = Tools::getValue('GA_TRACK_BACKOFFICE_ENABLED');
 
         if (!empty($gaAccountId)) {
             Configuration::updateValue('GA_ACCOUNT_ID', $gaAccountId);
@@ -217,8 +217,8 @@ class ConfigurationForm
         }
 
         if (null !== $gaTrackBackOffice) {
-            Configuration::updateValue('GA_TRACK_BACKOFFICE_DISABLED', (bool) $gaTrackBackOffice);
-            $treatmentResult .= $this->module->displayConfirmation($this->module->l('Settings for Disable Back Office tracking updated successfully'));
+            Configuration::updateValue('GA_TRACK_BACKOFFICE_ENABLED', (bool) $gaTrackBackOffice);
+            $treatmentResult .= $this->module->displayConfirmation($this->module->l('Settings for Enable Back Office tracking updated successfully'));
         }
 
         return $treatmentResult;
--- a/ps_googleanalytics.php
+++ b/ps_googleanalytics.php
@@ -139,7 +139,6 @@ class Ps_Googleanalytics extends Module
     {
         $gaTagHandler = new PrestaShop\Module\Ps_Googleanalytics\Handler\GanalyticsJsHandler($this, $this->context);
 
-        $output = '';
         $output = $gaTagHandler->generate(
             $this->context->cookie->__get('ga_admin_refund'),
             true
--- a/ps_googleanalytics.php
+++ b/ps_googleanalytics.php
@@ -139,11 +139,15 @@ class Ps_Googleanalytics extends Module
     {
         $gaTagHandler = new PrestaShop\Module\Ps_Googleanalytics\Handler\GanalyticsJsHandler($this, $this->context);
 
-        echo $gaTagHandler->generate(
+        $output = '';
+        $output = $gaTagHandler->generate(
             $this->context->cookie->__get('ga_admin_refund'),
             true
         );
         $this->context->cookie->__unset('ga_admin_refund');
+        $this->context->cookie->write();
+
+        return $output;
     }
 
     /**
--- a/classes/Hook/HookDisplayBackOfficeHeader.php
+++ b/classes/Hook/HookDisplayBackOfficeHeader.php
@@ -48,10 +48,6 @@ class HookDisplayBackOfficeHeader implements HookInterface
      */
     public function run()
     {
-        if (Configuration::get('GA_TRACK_BACKOFFICE_DISABLED')) {
-            return '';
-        }
-
         $js = '';
         if (strcmp(Tools::getValue('configure'), $this->module->name) === 0) {
             $this->context->controller->addCSS($this->module->getPathUri() . 'views/css/ganalytics.css');
--- a/classes/Hook/HookDisplayHeader.php
+++ b/classes/Hook/HookDisplayHeader.php
@@ -54,10 +54,6 @@ class HookDisplayHeader implements HookInterface
             return '';
         }
 
-        if (Configuration::get('GA_TRACK_BACKOFFICE_DISABLED') && $this->backOffice) {
-            return '';
-        }
-
         $this->context->controller->addJs($this->module->getPathUri() . 'views/js/GoogleAnalyticActionLib.js');
 
         $shops = Shop::getShops();
@@ -81,6 +77,7 @@ class HookDisplayHeader implements HookInterface
         $this->context->smarty->assign(
             [
                 'backOffice' => $this->backOffice,
+                'trackBackOfficeDisabled' => Configuration::get('GA_TRACK_BACKOFFICE_DISABLED'),
                 'currentShopId' => $currentShopId,
                 'userId' => $userId,
                 'gaAccountId' => Tools::safeOutput(Configuration::get('GA_ACCOUNT_ID')),
--- a/classes/Hook/HookDisplayHeader.php
+++ b/classes/Hook/HookDisplayHeader.php
@@ -77,7 +77,7 @@ class HookDisplayHeader implements HookInterface
         $this->context->smarty->assign(
             [
                 'backOffice' => $this->backOffice,
-                'trackBackOfficeDisabled' => Configuration::get('GA_TRACK_BACKOFFICE_DISABLED'),
+                'trackBackOffice' => Configuration::get('GA_TRACK_BACKOFFICE_ENABLED'),
                 'currentShopId' => $currentShopId,
                 'userId' => $userId,
                 'gaAccountId' => Tools::safeOutput(Configuration::get('GA_ACCOUNT_ID')),
--- a/views/templates/hook/ga_tag.tpl
+++ b/views/templates/hook/ga_tag.tpl
@@ -28,11 +28,3 @@
     </script>
     {/literal}
 {/if}
-
-{if ($jsState != 1 && $isBackoffice === true)}
-    {literal}
-    <script type="text/javascript">
-        ga('send', 'pageview');
-    </script>
-    {/literal}
-{/if}
--- a/views/templates/hook/ps_googleanalytics.tpl
+++ b/views/templates/hook/ps_googleanalytics.tpl
@@ -44,7 +44,7 @@
     {if $gaAnonymizeEnabled}
         ga('set', 'anonymizeIp', true);
     {/if}
-    {if $backOffice && $trackBackOfficeDisabled}
+    {if $backOffice && !$trackBackOffice}
         ga('set', 'nonInteraction', true);
     {else}
         ga('send', 'pageview');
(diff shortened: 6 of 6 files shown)


ps_googleanalytics.php
```diff
@@ -139,11 +139,15 @@ public function hookDisplayAdminOrder()
     {
         $gaTagHandler = new PrestaShop\Module\Ps_Googleanalytics\Handler\GanalyticsJsHandler($this, $this->context);
 
-        echo $gaTagHandler->generate(
+        $output = '';
```
kpodemski, line 142: @Hlavtox do we really need this empty var here?

Hlavtox: @kpodemski Maybe not?

If the function somehow fails, is it ok if we return false or whatever?

kpodemski:
Didn't test it but looks good for me 👍 

I'm not sure why we should allow pageViews in back-office in the first place?

ps_googleanalytics.php
kpodemski: @Hlavtox yes, unless there's a check `if (Hook::exec)` but that's not the case in PrestaShop from what I know :)

kpodemski: @Hlavtox you can check this by having `return false;` here and creating other module which is also hooked into `displayAdminOrder`, just to make sure ;)

Hlavtox: Removed ;-)

Progi1984:
Hi @Hlavtox, before going inside this PR, we need to fix this issue #74 and create specs for this module. If you are interested, you can open a PR on the repository [prestashop-specs](https://github.com/PrestaShop/prestashop-specs/tree/master/modules).

Hlavtox:
@Progi1984 I do not change any functionality. I restore the compatibility for other logic.

Hlavtox:
@Progi1984 Can you check the PR again? Why we need to wait a century until somebody writes a module spec when I do not change the the module behavior?

I only repair what somebody f***ed up. 🤷‍♂️

classes/Form/ConfigurationForm.php
```diff
@@ -176,7 +176,7 @@ public function generate()
         $helper->fields_value['GA_USERID_ENABLED'] = Configuration::get('GA_USERID_ENABLED');
         $helper->fields_value['GA_CROSSDOMAIN_ENABLED'] = Configuration::get('GA_CROSSDOMAIN_ENABLED');
         $helper->fields_value['GA_ANONYMIZE_ENABLED'] = Configuration::get('GA_ANONYMIZE_ENABLED');
-        $helper->fields_value['GA_TRACK_BACKOFFICE_DISABLED'] = Configuration::get('GA_TRACK_BACKOFFICE_DISABLED');
+        $helper->fields_value['GA_TRACK_BACKOFFICE_ENABLED'] = Configuration::get('GA_TRACK_BACKOFFICE_ENABLED');
```
eternoendless, line 179: Changing this settings requires an [upgrade script](https://devdocs.prestashop.com/1.7/modules/creation/enabling-auto-update/) so that settings from previous versions (`GA_TRACK_BACKOFFICE_DISABLED`) are cleaned up an carried over to `GA_TRACK_BACKOFFICE_ENABLED` in the new version.

eternoendless (changes requested):
Could you please explain why the GA script must be initialized even when you choose not to track in BO?

ttoine:
BTW, @Hlavtox, many people have contacted me about a couple of your recent posts, including one in this discussion.

So, could you please re-read the [Code of Conduct](https://github.com/PrestaShop/PrestaShop/blob/develop/CODE_OF_CONDUCT.md)? And then, try to adapt your tone, so discussions could be respectful and constructive, without aggressiveness? Because, if you continue like that, I guess that some people will fear to collaborate with you, even if your contribution to the project is interesting.

classes/Form/ConfigurationForm.php
kpodemski: Previous version was never released, is it needed in that case?

Hlavtox: @eternoendless The setting only made it to the dev version, no need to do an upgrade script I guess?

Hlavtox:
@eternoendless 

`Could you please explain why the GA script must be initialized even when you choose not to track in BO?`
Because there is a refund functionality in the module that outputs JS code in the middle of the page. Imagine something like `ga(action = 'refund', product = '123');`
In the way @matks has done it, this will stop working, because the JS code is killed before executing the header hook. BO tracking option should only affect the pageviews, not the entire script being blocked.

Hlavtox:
@ttoine Ok, let's go through it again.

- I am adding a order refund functionality to Google Analytics.
- This takes a century to review.
- In the meantime, you completely refactor the module and change everything.
- OK, I redo everything.
- Somebody messes up and the module suddenly tracks backoffice views.
- Somebody messes up even more and completely kills the JS code my PR relies on. When I tell him several times that it cannot be done that way, my comments and messages are ignored.
- OK, let's fix it.
- I make a PR with all the fixes and @Progi1984 that my PR needs to wait until you write some kind of module specifications? With the speed everything takes at Prestashop it will take a year. And what kind of specifications you need to write for Google Analytics module? Just do it as Google says.

ttoine:
I understand, I guess we all understand the frustration. Unfortunately, like in many other open source projects, this is happening here too. (not too often, thankfully). It happened before to other people in the past. And, it will happen again, for sure.

But being agressive and using insults will not help at all. It will just make people less motivated to collaborate with you.

classes/Form/ConfigurationForm.php
eternoendless: Oh in that case it's not needed. I thought it was already released.

eternoendless (approved):
Okay thanks for the explanation. Changes look good according to what you said. I'm not familiar enough with this module to understand the implications, WDYT @matks ?

Hlavtox:
@eternoendless Thank you! :-)

eternoendless:
BTW the reason @Progi1984 asks for specifications is that since we don't know how the module is expected to behave, it's very hard to validate changes. This is unfortunately the case in many places throughout the project because previous maintainers didn't leave a "paper trail", so to speak, to allow other people to pick up the job afterwards. So we're left with reverse engineering and guesswork, which obviously leads to many, may problems.

This is changing, but as you know already, it also takes a lot of time.

Hlavtox:
@eternoendless I perfectly understand it, but I would not hold off a fix that could upset a lot of people because of it. :-(

Imagine how many users will just click upgrade in their Prestashop BO and now they will have some mess in their statistics. Most visited page will now suddenly be BO orders... 😄

And regarding the specs, as I said, I think it's pretty easy in case of this module, isn't it. There is not much to change or alter, if the implementation is done per Google's documentation. [https://developers.google.com/analytics/devguides/collection/analyticsjs/enhanced-ecommerce#measuring-transactions](https://developers.google.com/analytics/devguides/collection/analyticsjs/enhanced-ecommerce#measuring-transactions)

Maybe some users want to do some non-standard things (send amount with taxes, where it's supposed to be without etc.), but that's their responsibility.

eternoendless:
> And regarding the specs, as I said, I think it's pretty easy in case of this module, isn't it

Great! I suppose that means you can write them! (just kidding, but if you did that it would be awesome 😄)

> And when I post some advice and somebody just completely ignores it, it just makes me think that the person thinks I'm stupid and incompetent

Oh no, rest assured, that's absolutely not the case. It's that we're DDoS by dozens of things every day and we can't invest as much time as we'd like in everything. That said, the fact that you're active in the project and that you point out nontrivial stuff has made you known to the team and, if anything, be _less_ "ignored" than others.

matks:
Hi @Hlavtox thank you for the detailed explanations about this PR

> # This PR fixes following bugs and confusions
>
> ### JS code that was killed before page output
> * JS code that is not included in a backoffice header because of #72. This PR changes the logic and disables only the pageview itself in the TPL.
> * This logic needs to work correctly because other functionality depends on it.
>   -- Default actionProductCancel functionality of the module
>   -- Order refund functionality I am adding here: #66

Approved, it's clean ✅ . If we get QA validation, it's a yes.

> ### Backoffice views enabled by default
> * Until 4.0.0, this module did not track backoffice views.
> * There was a bug in 4.0.0 that enabled backoffice pageviews to be sent.
> * @matks decided to keep the tracking there and disable it when wanted.
> * Based on previous behavior, the default state should be that the module should not track backoffice views.
> * If users upgrade to version 4, they would not be happy that it's behaving differently without their knowledge and they have to go to module settings to disable it.

This is discutable. v4.0.0 is a new version that does bring in new features. We can see, when looking at the code, that _this new 'tracking in BO' feature was wanted_.

The reason why I enabled the tracking in BO per default was to ensure that order cancel or refund events would be sent this way. If we follow your changes, indeed there is no need to enable tracking by default.

> ### Doubled and tripled pageviews
> * Removes a code that sends two or three extra pageviews in BO. This code is not present in v3. Pageview is sent in the header, no need to have it in the other TPL. This code was not present in v3 of the module.

I think you refer to the code statement in `views/templates/hook/ga_tag.tpl` right ?

Without better exploration of the code and [without documentation / specifications](https://github.com/PrestaShop/ps_googleanalytics/issues/74) I don't think we can simply remove this code. We cannot simply remove something without being 100% sure it is not useful. Else we might end up in [nasty situations](https://pbs.twimg.com/media/EirQkKhXYAM8A57?format=jpg&name=small). Moreover this is open source, people who use it are usually against removing things from the software.

We can explore this further to validate that this part of the code must go.

This is not something I write to annoy you. _We need to be extra careful when removing something_ because we don't know the inner workings of this module. As explained by @eternoendless because previous maintainers didn't leave a "paper trail", we're left with reverse engineering and guesswork, and this is not reliable.

So **if we can assert with a trust percentage high enough that this code should be removed** I'll approve the PR. I will not approve this part of the PR without this assertion. Once again, _extra caution is need here_. It's no fun, it's long, bothersome but this is the only safe way for a better release.

> ### Code output to order admin page was not working
> * Fixes an echo in one hook.
> * Echo does not work on migrated pages and needs to be changed to return.

Approved, it's clean ✅ . If we get QA validation, it's a yes.

## Summary of 'blocking items'

1. Need an exploration of the code and usecases to make sure code `views/templates/hook/ga_tag.tpl` can be removed
2. Need to make sure we're all OK with disabling tracking in Back-Office by default

matks:
@Hlavtox After checking with multiple stakeholders, I encountered no issues about putting the BO tracking on opt-in (= disabled by default ; enabled when merchant chooses to)

kpodemski:
I wonder what's the point of sending pageview from back-office, isn't it mistake in the first place?

matks:
> I wonder what's the point of sending pageview from back-office, isn't it mistake in the first place?

I wonder too, but if you look at this code:
https://github.com/PrestaShop/ps_googleanalytics/blob/master/views/templates/hook/ga_tag.tpl#L32
```
{if ($jsState != 1 && $isBackoffice === true)}
    {literal}
    <script type="text/javascript">
        ga('send', 'pageview');
    </script>
    {/literal}
{/if}
```

Unless it's very badly written, the intention is clearly "when we are in BO and this weird `jsState` is set to 1, send a pageview".

matks:
Ok, after digging into the git history I think I found the root cause

In commit https://github.com/PrestaShop/ps_googleanalytics/commit/3ec09fea2a1afa02c55402e5426eef4796c78d59#diff-74214ee2c66b22a15e9ec506a2e0b1af
the script is
```
{if ($jsState != 1 && $isBackoffice == 0)}
    {literal}
    <script type="text/javascript">
        ga('send', 'pageview');
    </script>
    {/literal}
{/if}
```
So we can see it's not intended for backoffice usage

But later commit https://github.com/PrestaShop/ps_googleanalytics/commit/07563209a788a151730fc2ff5f3a678d7af47bb9#diff-74214ee2c66b22a15e9ec506a2e0b1af
change
```
{if ($jsState != 1 && $isBackoffice == 0)}
```
into
```
{if ($jsState != 1 && $isBackoffice === true)}
```

kpodemski:
Oh, nice catch, so it was a mistake.

I think that even tho this module seems to be simple, having few points of how certain things in module should work is a must have... 

I'm pretty sure that this would be eventually highly appreciated in next refactoring of the module.

matks:
@Hlavtox Regarding this

> ### Doubled and tripled pageviews
> * Removes a code that sends two or three extra pageviews in BO. This code is not present in v3. Pageview is sent in the header, no need to have it in the other TPL. This code was not present in v3 of the module.

After digging through git history, we can now establish that this code did exist in v3 of the module. It was there https://github.com/PrestaShop/ps_googleanalytics/blob/3.1.3/ps_googleanalytics.php#L713 . It was code initially being built inside the main module file `ps_googleanalytics.php` and then inserted into the page.

However pageview is, in v4, being added through https://github.com/PrestaShop/ps_googleanalytics/blob/master/views/templates/hook/ps_googleanalytics.tpl which deprecates the usage of this file. This, combined with the condition inversion, results in the current situation.

I believe that these new elements allow us to assert with a trust percentage high enough that this code should be removed. In other words, I approve the Pull Request intent as these changes are both valid and desireable for the module behavior.

There is however something that might be wrong:

matks (approved):

Hlavtox:
@matks @kpodemski @eternoendless 

Guys, thank you for going through it and approving it!

_____________

Now, for you to have a calm sleep... :-D I think that on 99 %, the BO tracking was not intended. If you look on the following code. https://github.com/PrestaShop/ps_googleanalytics/blob/master/views/templates/hook/ps_googleanalytics.tpl

```
{if $backOffice}
  ga('set', 'nonInteraction', true);
{else}
  ga('send', 'pageview');
{/if}
```
The pageview was sent because $backoffice variable is not working, not because somebody wanted to be sent.

But, good thing in the end we fixed it and now it has yet another feature for people who would like to use it. 🚀

PierreRambaud (approved):

Progi1984 (approved):

Hlavtox:
bump pls, people are getting nonsense in their GA accounts @sarahdib

SD1982:
Hi @Hlavtox With this PR I have an error in the order view when i cancel a product
image

matks:
Hi @SD1982 I checked this error, I also have it with v3.3.0 and v4.0.0 this is not a regression. This is a compatibility issue between the module and PS 1.7.7.0. I think if you check on PS 1.7.6.9 it works.

SD1982:
Hi @Hlavtox as said [here](https://github.com/PrestaShop/ps_googleanalytics/pull/77#issuecomment-731644392) by matks and after more tests the PR seems ok but only for 1.7.6.9 for now.

Hlavtox:
@SD1982 Ok so you can merge it? The bug with cancelling products is not my problem I guess? 😛

matks:
Yes we can merge it 👍  because PS 1.7.6.9 is the current latest stable version

matks:
I added in the title the fact that it does support until PS 1.7.6

Compatibility issue for PS 1.7.7 is reported here https://github.com/PrestaShop/ps_googleanalytics/issues/84

Hlavtox:
Thanks @matks @SD1982 🚀

PR #79, opened 2020-09-21
Disable BO tracking by default

Hlavtox:
Until 4.0.0, this module did not track backoffice views.

During refactoring, somebody messed up and it tracked them.

@matks decided to keep the tracking there and disable it when wanted.

Based on previous behavior, the default state should be that the module should not track backoffice views. If users upgrade to version 4, they would not be happy that it's behaving differently without their knowledge.

### DIFF
--- a/classes/Form/ConfigurationForm.php
+++ b/classes/Form/ConfigurationForm.php
@@ -129,9 +129,9 @@ class ConfigurationForm
                 ],
                 [
                     'type' => 'switch',
-                    'label' => $this->module->l('Disable Back Office Tracking'),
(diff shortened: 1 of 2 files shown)


Progi1984:
@PrestaShop/product-team Are you ok for disabling BO tracking by default ?

LouiseBonnard:
Hi @matks, any reason why you decided to keep it?

Hlavtox:
@LouiseBonnard @matks Continue discussion here please https://github.com/PrestaShop/ps_googleanalytics/pull/77.

The changes are small but somehow tied together, so I decided to make a one combined PR.

PR #80, opened 2020-09-30
Clean file GoogleAnalyticActionLib.js from useless comments

prestaforum:
| Questions     | Answers
| Description?  | Please be specific when describing the PR. Every detail helps: What are you changing? Why?<br/>In this PR I am cleaning some code comments that I dont think bring any value. Either they provide obvious information that can be guessed from the code or they seem unrelated to the code logic (maybe some left-overs of previous versions ?). There is also dead commented out code that was kept for unknown reason as git is a better history state tracker than comments.
| Type?         | refacto
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | I only modify comments so I believe no tests are needed

### DIFF
--- a/views/js/GoogleAnalyticActionLib.js
+++ b/views/js/GoogleAnalyticActionLib.js
@@ -76,22 +76,19 @@ var GoogleAnalyticEnhancedECommerce = {
 	addToCart: function(Product) {
 		this.add(Product);
 		ga('ec:setAction', 'add');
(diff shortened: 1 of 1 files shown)


NeOMakinG (approved):
I agree, and looks like no QA needed

PierreRambaud (approved):

PierreRambaud:
Thank you @prestaforum

PR #81, opened 2020-10-01
Improve some comments in the code

prestaforum:
| Questions     | Answers
| Description?  | This PR suggests some rewording as some of the code comments are not always very well written.<br/>For example `hook home to display generate the product list associated to (...)` is a weird sentence and does not describe accurately the behavior of the function `hookDisplayHome()`.
| Type?         | improvement
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | I believe no test is needed as only code comments are modified


ps_googleanalytics.php
```diff
@@ -91,7 +91,7 @@ public function hookDisplayHeader($params, $back_office = false)
     }
 
     /**
-     * To track transactions
+     * Confirmation page hook ; to track transactions
```
PierreRambaud, line 94: Why adding `;` in these comments? :thinking:

prestaforum: My intent with this semi-colon was to show to reader, in a better way, the 2 parts of this comment: the 1st part explains what page is related to this hook (Checkout page ? Homepage ? Order confirmation page ?) ; the 2nd part explains why we register on this hook ; is it to send a transaction to Google Analytics, is it to send a pageview ...

Some hook names do not need further explanation so 1st part might be useless. Other hook names are sometimes confusing so the 1st part is useful.
However to keep the code homogenous I always add the 1st part separated by a semi-colon then the 2nd part. I believe uniformity helps the developer to picture the php class.

So I wanted to use a separator, but if `;` is not a valid one, any separator can do. What about `|` or `-` ?

PierreRambaud: No, it's ok, you explain it very well.  But I think separators can be confusing.
Here are some suggestions.

```suggestion
     * Confirmation page hook: to track transactions
```
```suggestion
     * Confirmation page hook.
     * This function is run to track transactions.
```

prestaforum: Good evening @PierreRambaud I apologize for not answering earlier, I have been very busy this last month.

I improve my PR to follow your suggestion.

PierreRambaud (approved):

PierreRambaud:
Thanks @prestaforum

PR #99, opened 2021-08-31
Prevent error when GoogleAnalyticEnhancedECommerce is not set

camlafit:
In some case GoogleAnalyticEnhancedECommerce is not set.
To prevent any javascript error as ```Uncaught ReferenceError: GoogleAnalyticEnhancedECommerce is not defined``` provide a test before to use it

| Questions         | Answers
| Description?      | Prevent Uncaught ReferenceError and block javascript execution
| Type?             | bug fix
| BC breaks?        |  no
| Deprecations?     |  no
| Fixed ticket?     |  Fixes https://github.com/PrestaShop/PrestaShop/issues/26093
| How to test?      | 
| Possible impacts? | Any

### DIFF
--- a/views/templates/hook/ga_tag.tpl
+++ b/views/templates/hook/ga_tag.tpl
@@ -21,9 +21,11 @@
     {literal}
     <script type="text/javascript">
         document.addEventListener('DOMContentLoaded', function() {
-            var MBG = GoogleAnalyticEnhancedECommerce;
-            MBG.setCurrency('{/literal}{$isoCode|escape:'htmlall':'UTF-8'}{literal}');
-            {/literal}{$jsCode nofilter}{literal}
+            if (typeof GoogleAnalyticEnhancedECommerce !== 'undefined') {
+                var MBG = GoogleAnalyticEnhancedECommerce;
+                MBG.setCurrency('{/literal}{$isoCode|escape:'htmlall':'UTF-8'}{literal}');
+                {/literal}{$jsCode nofilter}{literal}
+            }
         });
     </script>
     {/literal}


matks (approved):

matks (commented):
Thanks for the PR @camlafit !

To help the QA team explore and verify this PR, would it be possible to give us a little more details about how to trigger the error you mention?

I understand you need a GA account without GoogleAnalyticEnhancedECommerce on. But then what do you do? Add some products in cart and checkout? Which webpage will show the error?

The best would be to open a bug report with these data 😉  that would be very helpful for us to track and explore the matter.

camlafit:
Hello

thanks for your return @matks , I've just opened #100 to provide more information.
But in any case looks a good practice to check any js variable before to use it :)

As disposition to provide more information

atomiix (approved):

Progi1984 (approved):

khouloudbelguith:
Hi @camlafit,

Sorry, I tried again with PS1780 and ps_googleanalytics v4.1.0, but I did not manage to reproduce the issue.
I attached a screen record
https://watch.screencastify.com/v/eVg2WiLAD5QTkpnmsmNN

Thanks!

camlafit:
Hello

I can only confirm initial behavior. It's took from a real case in production. Without patch error occur.
And I'm sorry to repeat, I don't why you want refuse this patch. We improve quality code, solve a real use case and don't break anything. 
At my level looks as a simple good practice 

https://user-images.githubusercontent.com/209915/140529080-bd31af66-909a-44b2-8721-8eec8713e104.mp4

matks:
@camlafit Some explanations: we cannot validate a contribution that we cannot reproduce ourselves. We have the responsibility to verify every line of code that is added to the project. We cannot verify that a line submitted by you will make an issue go away if we cannot reproduce the issue.

it does not mean the issue you mention does not exist ; it means that in order to observe it some specific steps or setting are needed.

also having this situation described would help us verify that it does not only get fixed now, we can also verify in the future, following new changes in the project, that it is still corrected, we dont want a bug to be fixed then come back

So this is about "why we need to reproduce something" 😄 

Back to this PR, even though we dont have a clear scenario I think indeed the code changes are simple enough and make sense even if we dont know how it could fail currently. I think we could merge them, wdyt @PrestaShop/prestashop-maintainers  ?

atomiix:
@matks AFAIK this code is pretty safe to be merged!

matks:
Thank you @camlafit

PR #104, opened 2021-11-15
Update sent transaction date with type sql NOW()

gadnis:
`['date_add' => 'NOW()']` does not update the `date_add` column value

### DIFF
--- a/classes/Hook/HookDisplayBackOfficeHeader.php
+++ b/classes/Hook/HookDisplayBackOfficeHeader.php
@@ -90,7 +90,7 @@ class HookDisplayBackOfficeHeader implements HookInterface
                             if (!empty($transaction)) {
                                 $ganalyticsRepository->updateData(
                                     [
-                                        'date_add' => 'NOW()',
+                                        'date_add' => ['value' => 'NOW()', 'type' => 'sql'],
                                         'sent' => 1,
                                     ],
                                     'id_order = ' . (int) $row['id_order'] . ' AND id_shop = ' . (int) $this->context->shop->id
--- a/controllers/admin/AdminGanalyticsAjax.php
+++ b/controllers/admin/AdminGanalyticsAjax.php
@@ -33,7 +33,7 @@ class AdminGanalyticsAjaxController extends ModuleAdminController
             (new GanalyticsRepository())->updateData(
                 [
                     'sent' => 1,
-                    'date_add' => 'NOW()',
+                    'date_add' => ['value' => 'NOW()', 'type' => 'sql'],
                 ],
                 'id_order = ' . $orderId
             );


matks:
This PR is similar to https://github.com/PrestaShop/ps_googleanalytics/pull/68 

@gadnis can you provide a test case where, before your PR it fails, and after your PR it is successfull?

gadnis:
@matks I'm not strong in testing, sorry..

gadnis:
@matks but maybe you could tell me how to run phpstan test for this module? I read [php-dev-tools readme](https://github.com/PrestaShop/php-dev-tools/blob/master/README.md) and I understand that this module has everything set up but not sure how to run in for this module. I would love to understand how test works with php-dev-tools package. Could you give me a link where to find better explanation how to run tests for this module or at least write me a command line for example test?

matks:
@gadnis I was not thinking about automated tests 😄  just a scenario (= steps) that, as a human, I can follow to reproduce faulty behavior and confirm that it does not trigger anymore after the patch is applied

Progi1984 (approved):

PierreRambaud (approved):

Progi1984:
@gadnis Thank you for your contribution. :hugs: 

Could you give us a scenario which permits to test the bug before and after this PR ?

Thanks :+1:

PierreRambaud:
@matks do you remember what was the original issue? I'm not able to retrieve it regarding the PR you did long time ago.
The issue is the same

matks:
> #68

I wrote this statement 18 months ago

>  When placing an order, the timestamp data was badly registered in MySQL (0000-00-00 00:00:00 instead of current datetime) inside ps_ganalytics table. It's because NOW() is specific value and must be configured as such, else PrestaShop considers it's the string NOW() and not the operator.

So looks like a bad interpretation of NOW by PrestaShop ORM. Basically PrestaShop ORM would try to record the string "NOW" inside MySQL inside of the function "NOW()".

PierreRambaud:
> I wrote this statement 18 months ago => #68
> 
> > When placing an order, the timestamp data was badly registered in MySQL (0000-00-00 00:00:00 instead of current datetime) inside ps_ganalytics table. It's because NOW() is specific value and must be configured as such, else PrestaShop considers it's the string NOW() and not the operator.
> 
> So looks like a bad interpretation of NOW by PrestaShop ORM. Basically PrestaShop ORM would try to record the string "NOW" inside MySQL inside of the function "NOW()".

Yes that's exaclty this. Could we do the QA in our side as it's technical?

matks:
I think yes

atomiix:
Before the PR:
Capture d’écran 2021-12-24 à 12 21 24

After the PR:
Capture d’écran 2021-12-24 à 12 22 41

This PR is QA ✔️
Thanks @gadnis

PR #107, opened 2021-12-24
Check if array key products exists before using it

atomiix:
| Questions         | Answers
| Description?      | An array key not always available was access without checking if it actually exists, leading to an exception in php >= 7.4. This PR fixes by checking it and returning an empty array if it doesn't exist.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes PrestaShop/PrestaShop#26539
| How to test?      | Please see PrestaShop/PrestaShop#26539
| Possible impacts? |


PierreRambaud (approved):

PierreRambaud:
Thank you @atomiix

PR #116, opened 2022-04-04
Moved some code from footer to head tag

Progi1984:
| Questions         | Answers
| ------------------| -------------------------------------------------------
| Description?      | Moved some code from footer to head tag
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes PrestaShop/PrestaShop#28004
| How to test?      | Cf. PrestaShop/PrestaShop#28004

PR #117, opened 2022-06-25
Fix deprecated CHARSET utf8 in Install and Upgrade

leemyongpakvn:
ps_analytics_deprecated_charset_utf8

| Questions         | Answers
| Description?      | You should also be aware that the utf8mb3 character set is deprecated and you should expect it to be removed in a future MySQL release. Please use utf8mb4 instead.
| BC breaks?        | no
| Deprecations?     | yes
| Fixed ticket?     | Fixes PrestaShop/PrestaShop#28857
| How to test?      | Check ganalytics and ganalytics_data tables before PR with phpmyadmin you will see Collation = ---

### DIFF
--- a/classes/Database/Install.php
+++ b/classes/Database/Install.php
@@ -61,14 +61,14 @@ class Install
             PRIMARY KEY (`id_google_analytics`),
             KEY `id_order` (`id_order`),
             KEY `sent` (`sent`)
-        ) ENGINE=' . _MYSQL_ENGINE_ . ' DEFAULT CHARSET=utf8 AUTO_INCREMENT=1';
+        ) ENGINE=' . _MYSQL_ENGINE_ . ' DEFAULT CHARSET=utf8mb4 AUTO_INCREMENT=1';
 
         $sql[] = 'CREATE TABLE IF NOT EXISTS `' . _DB_PREFIX_ . 'ganalytics_data` (
             `id_cart` int(11) NOT NULL,
             `id_shop` int(11) NOT NULL,
             `data` TEXT DEFAULT NULL,
             PRIMARY KEY (`id_cart`)
-        ) ENGINE=' . _MYSQL_ENGINE_ . ' DEFAULT CHARSET=utf8';
+        ) ENGINE=' . _MYSQL_ENGINE_ . ' DEFAULT CHARSET=utf8mb4';
 
         foreach ($sql as $query) {
             if (!Db::getInstance()->execute($query)) {
--- a/upgrade/upgrade-4.1.3.php
+++ b/upgrade/upgrade-4.1.3.php
@@ -26,6 +26,7 @@ if (!defined('_PS_VERSION_')) {
  */
 function upgrade_module_4_1_3($object)
 {
-    return $object->unregisterHook('displayHome') 
-        && Db::getInstance()->execute('ALTER TABLE `' . _DB_PREFIX_ . 'ganalytics` CHARACTER SET = utf8mb4 COLLATE utf8mb4_general_ci;');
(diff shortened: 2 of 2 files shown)


okom3pom:
Hello @leemyongpakvn. 

Thank you for your Pull Request what do you think also to add an upgrade file to change also the charset of the old version of the module.

upgrade/upgrade-4.1.3.php
```diff
@@ -26,5 +26,6 @@
  */
 function upgrade_module_4_1_3($object)
 {
-    return $object->unregisterHook('displayHome');
+    return $object->unregisterHook('displayHome')
```
okom3pom, line 29: You can remove $object->unregisterHook('displayHome')

okom3pom (changes requested):

upgrade/upgrade-4.1.3.php
leemyongpakvn: @okom3pom It is a part of @Progi1984's PR 116 https://github.com/PrestaShop/ps_googleanalytics/pull/116/files#diff-6b48dcebdefb30591797f7fcb3b79fd86e2a6769c105d610b9fd32e787ce971e

okom3pom: You are right sorry

okom3pom (approved):

classes/Database/Install.php
```diff
@@ -61,14 +61,14 @@ public function installTables()
             PRIMARY KEY (`id_google_analytics`),
             KEY `id_order` (`id_order`),
             KEY `sent` (`sent`)
-        ) ENGINE=' . _MYSQL_ENGINE_ . ' DEFAULT CHARSET=utf8 AUTO_INCREMENT=1';
+        ) ENGINE=' . _MYSQL_ENGINE_ . ' DEFAULT CHARSET=utf8mb4 AUTO_INCREMENT=1';
```
matks, line 64: This is a good thing 👍

upgrade/upgrade-4.1.3.php
```diff
@@ -26,5 +26,7 @@
...
+        && Db::getInstance()->execute('ALTER TABLE `' . _DB_PREFIX_ . 'ganalytics` CHARACTER SET = utf8mb4 COLLATE utf8mb4_general_ci;')
+        && Db::getInstance()->execute('ALTER TABLE `' . _DB_PREFIX_ . 'ganalytics_data` CHARACTER SET = utf8mb4 COLLATE utf8mb4_general_ci;');
```
matks, line 31: I wonder about this. Should we force the charset? Or should we not do something?

- so people who install ps_googleanalytics for the 1st time will have utf8mb4
- people who upgrade to 4.1.3 will keep their previous charset

I think maybe some people already fixed their charset in database or even chose another one. So forcing, during the upgrade step, the charset is a risky move.

leemyongpakvn: Agree. We need keep changes as safe as possible.

okom3pom: Hello @matks 

We could check the character set before?
This big change was made in 1.7.7.0 without any verification, this is the recommendation of the Mysql documentation 

https://github.com/PrestaShop/autoupgrade/blob/dev/upgrade/sql/1.7.7.0.sql#L64-L307

leemyongpakvn:
Maybe, Dev team will add this kind of task into autograde module. Otherwise, we will do it in another PR.

kpodemski:
Hi @leemyongpakvn 

Thanks for your PR. I think upgrades for modules should be inside modules, not an `autoupgrade` module. Your PR is valid.

PR #121, opened 2022-06-30
Display Cross-Domain confirmation only when Multistore Enabled

leemyongpakvn:
In `function treat` that process ConfigurationForm save at https://github.com/PrestaShop/ps_googleanalytics/blob/dev/classes/Form/ConfigurationForm.php#L205-L229
If `GA_CROSSDOMAIN_ENABLED` is not set, `Tools::getValue('GA_CROSSDOMAIN_ENABLED')` will return `False`, then `$gaCrossdomainEnabled` is assigned to `False`.
In turn `(null !== $gaCrossdomainEnabled)` is `True` because `(null !== False)` is `True`.
That's why ConfigurationForm Save display CrossDomain confirmation even when multistore is `Disabled`. Furthermore, CrossDomain setting reused UserIDTracking's Translation string, not its own.
ps_googleanalytics_multishop_enabled

| Questions         | Answers
| Description?      | Solution: calculate and use `$is_multistore_active` like `function generate `did at line 49 and 169, add a new Translation string also. 
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes PrestaShop/PrestaShop/issues/28005
| How to test?      | After this PR, with multistore `Disabled` the wrong confirmation described in issue 28005 will disappear, with multistore `Enabled` and more than 1 shop the right confirmation will display like above screenshot.

### DIFF
--- a/classes/Form/ConfigurationForm.php
+++ b/classes/Form/ConfigurationForm.php
@@ -204,6 +204,9 @@ class ConfigurationForm
      */
     public function treat()
     {
+        // Check if multistore is active
+        $is_multistore_active = Shop::isFeatureActive();        
+        
         $treatmentResult = '';
         $gaAccountId = Tools::getValue('GA_ACCOUNT_ID');
         $gaUserIdEnabled = Tools::getValue('GA_USERID_ENABLED');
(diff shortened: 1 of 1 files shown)


leemyongpakvn:
CS Fixer No 2 not PR2

matks (commented):
@leemyongpakvn sorry for late answer

what happens for single-store users? using your PR they cannot use cross domain

leemyongpakvn:
@matks  **Nothing** change for single-store users **after** applying this PR. They cannot use cross domain **before** applying this PR. It is the **current logic**. Please check [ConfigurationForm.php#L49](https://github.com/PrestaShop/ps_googleanalytics/blob/dev/classes/Form/ConfigurationForm.php#L49) and [ConfigurationForm.php#L169-L187](https://github.com/PrestaShop/ps_googleanalytics/blob/dev/classes/Form/ConfigurationForm.php#L169-L187) for details.
Below is what single-store users see in BO > Modules > ps_googeanalytics > Configure before PR. You can can see the same screen is the related issue's video clip. (No cross-domain setting for single-store users)
ps_googleanalytics_config_for_singleStore

leemyongpakvn:
close as conflict with master branch, use PR 126 instead.

PR #122, opened 2022-07-05
Remove redundant function in class Hook

leemyongpakvn:
| Questions         | Answers
| Description?      | 2 functions to do a simple assignment is unnecessary.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes PrestaShop/PrestaShop/issues/28905
| How to test?      | Please check Files changed and the above issue.

### DIFF
--- a/classes/Hook/HookDisplayHeader.php
+++ b/classes/Hook/HookDisplayHeader.php
@@ -155,14 +155,6 @@ class HookDisplayHeader implements HookInterface
      */
     public function setBackOffice($backOffice)
     {
-        $this->acknowledgeBackOfficeContext($backOffice);
-    }
-
-    /**
-     * @param bool $isBackOffice
-     */
-    public function acknowledgeBackOfficeContext($isBackOffice)
-    {
-        $this->backOffice = $isBackOffice;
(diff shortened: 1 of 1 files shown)


matks (changes requested):
Hello @leemyongpakvn removing a function is a Breaking Change. You're right this code is useless. But we need to preserve the Public API (see below references).

As I see no clear benefit in removing this, but instead it might be a little bother for modules that rely on `acknowledgeBackOfficeContext()` I suggest we keep it 😉 

### References

https://stackoverflow.com/questions/9463613/what-does-public-api-mean-in-semantic-versioning?rq=1#:~:text=E.g.%2C%20if%20you're%20developing,valid%20on%20a%20later%20version.

https://github.com/PrestaShop/ADR/blob/a255300fe2b47a05f694f10024d9e5cf8be8a842/0017-backward-compatibility-promise.md

leemyongpakvn:
Agree. I don't like Breaking Change either.
But I'm quite sure that `acknowledgeBackOfficeContext()` is not used any elsewhere - `grep` in the whole Prestashop directory but not found.

PR #123, opened 2022-07-05
Remove redundant function in classes Handler

leemyongpakvn:
| Questions         | Answers
| Description?      | function **isModuleEnabled** is redundant since v4.1.0 and need to be removed
| BC breaks?        | no
| Deprecations?     | yes
| Fixed ticket?     | Fixes PrestaShop/PrestaShop/issues/28953
| How to test?      | Please check Files changed and the above issue.

### DIFF
--- a/classes/Handler/ModuleHandler.php
+++ b/classes/Handler/ModuleHandler.php
@@ -26,30 +26,6 @@ class ModuleHandler
 {
     /**
      * @param string $moduleName
-     *
-     * @return bool
-     */
-    public function isModuleEnabled($moduleName)
-    {
-        $module = Module::getInstanceByName($moduleName);
-
-        if (!($module instanceof Module)) {
-            return false;
-        }
-
-        if (false === Module::isInstalled($moduleName)) {
-            return false;
-        }
-
-        if (false === $module->active) {
-            return false;
-        }
-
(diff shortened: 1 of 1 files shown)


matks (changes requested):
Hello @leemyongpakvn same answer as https://github.com/PrestaShop/ps_googleanalytics/pull/122#pullrequestreview-1031710728:

 Removing a function is a Backward Incompatible Change. You're right this code is useless. But we need to preserve the Public API (see below references).

In open source, because of this backward compatibility constraint, we quite often have to keep a lot of old code around because of this.

What we can do however is [deprecate](https://www.techopedia.com/definition/24828/deprecated#:~:text=Techopedia%20Explains%20Deprecated-,What%20Does%20Deprecated%20Mean%3F,be%20used%20in%20the%20meantime.) code to mark it as "don't use it anymore"

### References

https://stackoverflow.com/questions/9463613/what-does-public-api-mean-in-semantic-versioning?rq=1#:~:text=E.g.%2C%20if%20you're%20developing,valid%20on%20a%20later%20version.

https://github.com/PrestaShop/ADR/blob/a255300fe2b47a05f694f10024d9e5cf8be8a842/0017-backward-compatibility-promise.md

PR #124, opened 2022-09-01
Add missing domains and use trans instead of l

jolelievre:
| Questions         | Answers
| ------------------| -------------------------------------------------------
| Description?      | Add missing domains and use trans instead of l
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | ~
| How to test?      | ~
| Possible impacts? | ~

PR #127, opened 2022-12-02
shorten trans call

leemyongpakvn:
| Questions     | Answers
| Description?  | Module class has its own [trans](https://github.com/PrestaShop/PrestaShop/blob/develop/classes/module/Module.php#L3363) function so we need to replace `getTranslator()->trans` by `trans` to keep consistency and briefness.
| Type?         | improvement
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes PrestaShop/Prestashop#30193
| How to test?  | Nothing changes in module's behavior. Check the Module `trans` code in the above Description is the best way verify that this PR is correct.

### DIFF
--- a/classes/Form/ConfigurationForm.php
+++ b/classes/Form/ConfigurationForm.php
@@ -70,13 +70,13 @@ class ConfigurationForm
         $helper->submit_action = 'submit' . $this->module->name;
         $helper->toolbar_btn = [
             'save' => [
-                'desc' => $this->module->getTranslator()->trans('Save', [], 'Modules.GAnalytics.Admin'),
+                'desc' => $this->module->trans('Save', [], 'Modules.GAnalytics.Admin'),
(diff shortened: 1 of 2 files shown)


mflasquin (approved):

MhiriFaten (commented):
Hello @leemyongpakvn ,

Following your code modification, the module gives an exception during its installation and accessing the configuration.

Here is a screenshot 

image

Could you check it please ?

leemyongpakvn:
@jolelievre I'm trying to make trans call consitent throughout the project and get a problem with `protected function trans` you added in [PR124](https://github.com/PrestaShop/ps_googleanalytics/pull/124/files) such as @MhiriFaten reported above. 
I think we can change to `public function trans` in this module OR move your code to [Module class](https://github.com/PrestaShop/PrestaShop/blob/8.0.0/classes/module/Module.php#L3363-L3368). Wdyt?

Hlavtox:
Thanks @MhiriFaten @leemyongpakvn!

PR #129, opened 2023-01-17
Update ps_googleanalytics.tpl

pitylee:
Update the tag for the new G- tag that utilizes tag manager for Analytics.

| Questions         | Answers
| Description?      |  In the future only this is used; every existing google analytics property will continue working - newly registered properties need this change.
| Type?             | improvement / new feature  / critical
| BC breaks?        | no
| Deprecations?     | yes 
| How to test?      | Check with an analytics id that starts with G- and one that starts with UA-
| Possible impacts? | No impact.

### DIFF
--- a/views/templates/hook/ps_googleanalytics.tpl
+++ b/views/templates/hook/ps_googleanalytics.tpl
@@ -17,6 +17,19 @@
  * International Registered Trademark & Property of PrestaShop SA
  *}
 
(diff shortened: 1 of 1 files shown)


Hlavtox (changes requested):
GA4 is already in development, and this is not enough to implement it.

pitylee:
Okay, waiting for the new version for some time, in the meantime I use this template on prod.

Thank you, I close the PR.

PR #141, opened 2023-03-24
Enable Backoffice Tracking Toggler not working

PrestaConnect:
Enable Backoffice Tracking Toggler not working

| Questions         | Answers
| Description?      | Please be specific when describing this change. What did you change? Why?
| Type?             | bug fix / improvement / new feature / refactor
| BC breaks?        | yes / no
| Deprecations?     | yes / no
| Fixed ticket?     | Fixes #{issue URL here}, Fixes #{another issue URL here}
| Sponsor company   | Your company or customer's name goes here (if applicable).
| How to test?      | Indicate how to verify that this change works as expected.

### DIFF
--- a/ps_googleanalytics.php
+++ b/ps_googleanalytics.php
@@ -178,9 +178,12 @@ class Ps_Googleanalytics extends Module
      */
     public function hookActionProductCancel($params)
     {
-        $hook = new PrestaShop\Module\Ps_Googleanalytics\Hooks\HookActionProductCancel($this, $this->context);
-        $hook->setParams($params);
-        $hook->run();
+        if (Configuration::get('GA_TRACK_BACKOFFICE_ENABLED')) {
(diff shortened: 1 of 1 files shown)


Hlavtox (changes requested):
@PrestaConnect This switch is not switching refunds and cancels, but page visits.

PrestaConnect:
> @PrestaConnect This switch is not switching refunds and cancels, but page visits.

But 

> @PrestaConnect This switch is not switching refunds and cancels, but page visits.

However, the analytics script appears on all pages in the back office. Mostly, analytics tracking is not a desirable situation for the back office in general. It's reasonable to run hooks for returns/cancellation or payment confirmation.

Hlavtox:
@PrestaConnect It appears, but it's not tracking because of this logic ;-)

`{if $backOffice && !$trackBackOffice}, 'non_interaction': true, 'send_page_view': false{/if}`

PR #130, opened 2023-02-07
GA4 fixes and improvements

Hlavtox:
| Questions         | Answers
| ------------------| -------------------------------------------------------
| Description?      | See below
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | -
| Sponsor company   | -
| How to test?      | Check that BO looks OK, use https://tagassistant.google.com/ to check that events are sent properly.

### Improvements and bugs fixed
- Improved backoffice layout and it's responsivity.
- Removed obsolete unused CSS file.
- Switched settings order so GA4 switch is on the top, before the ID.
- Added input helpers to some settings.
- Fixed - `page_view` event was sent in BO, even though the setting was disabled.
- Fixed event callback was not implemented. It's was not confirming sent orders in database.
- Fixed additional use of context currency in refunds, not needed, only if value is provided. And it must be orders currency, not context.
- Fixed extra condition that prevented revisiting confirmation if it failed. (No double submit don't worry.)
- Unified `purchase` event quotation marks.
- Fixed missing value, tax, shipping, affiliation to FO `purchase` event.
- Use json_encode to safely generate data, exluding callbacks.
- Fix `purchase` in FO - Missing order currency.
- Fix `select_item` - Wrong item format, price MUST BE without currency.
- Fix `add_to_cart` - Wrong item format, every strings should be original
- Fix `remove_from_cart` - Wrong item format, every strings should be original
- Fix `purchase` - Wrong item format, every strings should be original
- Fix - Begin_checkout not firing (footer was not reachable in checkout in 1.7) (item list not implemented)

### Improvements to do
- `add_payment_info` - Must use order currency, not context.
- `purchase` in BO - Must use order currency, not context.
- View_item_list not implemented
- Item list missing in BO purchase event
- View_cart not implemented
- Test multishipping, I think it would send only first order from the cart. Also, it should resolve a list of orders and send them using entries from order table, not cart.

### Screenshot
analytics

### If you want to use UA and GA4 together, install a copy of a module I created
[ps_oldgoogleanalytics.zip](https://github.com/PrestaShop/ps_googleanalytics/files/10824107/ps_oldgoogleanalytics.zip)

PR #144, opened 2023-04-01
Item data is not sent consistently to GA4 (wip)

dkarvounaris:
| Questions         | Answers
| Description?      | Inconsistent data submission is resulting in incomplete data, as also lost associations between data, in Google Analytics 4 reports. Details see below.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| How to test?      | TBD

# The issue and it's effects

There are inconsistencies in the data that are sent with some events, resulting in missing data in Google Analytics 4 reports. Additionally, the association between data in different columns is lost, making it difficult to interpret the numbers. Specifically:

In the following report, we switched to view "Items viewed", "Items added to cart" and "Items purchased" by brand. However, 100% the "Items viewed" are counted as one single brand with an empty name, and all purchases are counted as one entry with the brand reported as "(not set)". The only properly sectioned column is "Items added to cart," with the numbers having no longer a relation to the other columns. Additionally, as a result any actual brand reported in "Item revenue" has a value of 0, as all revenue is reported under "(not set)".

image

We do see something similar when the report is viewed by "Item category":

image

Although other reports may also be affected, the issue is most clearly observed in the "Monetization > Ecommerce purchases" report when displayed by "Item brand" or "Item category." The same report, listed by "Item name" does not show this issue, because "Item name" is a required parameter and sent everywhere. 

# What causes the issue?

The Google Analytics module is not submitting item data consistently, resulting in the discrepancies described above. The submitted data can be confirmed and analyzed in greater detail using the Google Tag Assistant debugger.

On the event "view_item", the brand is "[]", item category is "Graphics Cards":
image

On the event "select_item", the brand is an empty string, item category is "graphics-cards":
image

On the event "add_to_cart", the brand is "MSI" and item category is "graphics-cards":
image

On the event "purchase", the brand and item category are not reported to GA4:
image

index = item_id ???
image

cart value is NOT item price, should be quantity * item price?
image

As seen in the screenshots, the data may vary depending on the code which populates the `items` parameter for $eventData. Despite the fact that the `items` parameter is for all events always identical according to https://developers.google.com/tag-platform/gtagjs/reference/events, the Google Analytics module uses duplicated code in multiple locations and files, resulting in the incosistency of data, as well sometimes adding more or less data for the item.

(to be continued - commits in progress)


Hlavtox:
@dkarvounaris Hey, I can confirm there are maaaaany things still wrong and not implemented. I fixed many things in my PR here - https://github.com/PrestaShop/ps_googleanalytics/pull/130, but it's far not finished. So I am glad that you took the job! ❤️

I wanted to do it “properly” when Universal Analytics dies, so we can clean the module and remove all those conditions.

I will check and test this later, just make sure to also test that the changes didn't break UA. There are some common methods shared for both, but UA uses different format than GA4.

dkarvounaris:
@Hlavtox Yes I checked yesterday after posting this your PR too. At least one thing (missing item list) is definetely one I was going to solve with this. Don't look at the commits yet, I'll change them and they are far from finished (why it's all a draft at the moment). I noticed, the description is gonna be come also too long, I'll move most details into a separate comment and leave an overview in the main one, I think it would be better this way, no?

Hlavtox:
@Tofandel I fixed lot of things on dev branch. Now orders are nearly finished - https://github.com/PrestaShop/ps_googleanalytics/pull/150

Then we can also fix begin checkout and other events. :-)

Hlavtox:
@dkarvounaris This will be completely fixed by this PR ;-) https://github.com/PrestaShop/ps_googleanalytics/pull/153

Hlavtox:
Closing as all the issues should be fixed. Please try 5.0.0 version and let me know if you find any more tweaks to do. :-)

PR #146, opened 2023-04-04
Release 4.2.2

lartist:
| Questions         | Answers
| Description?      | Release 4.2.2
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | 
| Sponsor company   | 
| How to test?      | Test the whole module

### DIFF
--- a/classes/Form/ConfigurationForm.php
+++ b/classes/Form/ConfigurationForm.php
@@ -111,7 +111,7 @@ class ConfigurationForm
                     'name' => 'GA_ACCOUNT_ID',
                     'size' => 20,
                     'required' => true,
-                    'desc' => $this->module->trans('This information is available in your Google Analytics account. GA4 tracking ID starts with "G-", Universal Analytics with "UA-."', [], 'Modules.GAnalytics.Admin'),
+                    'desc' => $this->module->trans('This information is available in your Google Analytics account. GA4 tracking ID starts with "G-", Universal Analytics with "UA-".', [], 'Modules.GAnalytics.Admin'),
                 ],
                 [
                     'type' => 'switch',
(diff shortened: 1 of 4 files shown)


0x346e3730 (approved):

matthieu-rolland (commented):
phpstan failing, LGTM otherwise

Hlavtox (changes requested):
Wording typo, quotes after dot.

Is it really necessary to release and test whole module because one wording change?

lartist:
>Wording typo, quotes after dot.

Hi @Hlavtox, thanks for reporting the typo. I'm going to change it.

>Is it really necessary to release and test whole module because one wording change?

I totally agree with you, releasing and testing the whole module is a bit huge for a wording change. Unfortunately we have to. The 8.1.0 is coming soon and we have to extract the catalog of wordings. The way the extraction works is the following:

PrestaShop/PrestaShop is downloaded by [TranslationTool](https://github.com/PrestaShopCorp/TranslationTool) every built-in modules are installed with composer and then the extraction processes on the CORE AND the modules.

If we want an exhaustive catalog we have to release this version and if it occurs on another module we will do the same thing, even if it's for only one wording.

Maybe the extraction process can be improved. If you have any idea feel free to suggest it ;)

lartist:
Fix of typo here https://github.com/PrestaShop/ps_googleanalytics/pull/147

Hlavtox (approved):

AureRita (approved):
Hi @lartist 

Thank you for your PR, I tested it and it seems to works as expected, as you can see : 

image

image

matthieu-rolland (approved):
Can we merge or should we wait @lartist ?

PR #148, opened 2023-05-18
Remove old code, fix translation domains, bump compatibility

Hlavtox:
| Questions         | Answers
| Description?      | See below
| Type?             | bug fix
| BC breaks?        | yes
| Deprecations?     | no
| Fixed ticket?     | Fixes #https://github.com/PrestaShop/PrestaShop/issues/32175
| Sponsor company   | 
| How to test?      | 

### Description
- Dropped universal analytics support. **DO NOT RELEASE THIS MODULE BEFORE 1.7.2023!**
- Dropped legacy translation system and migrated to the new one.
- Fixes translation domain to allow proper translations and export.
- Removes compatibility with 1.6 (this was the last module) and bumps compatibility to 1.7.6 (first version with the new price format system).


config.xml
```diff
@@ -2,7 +2,7 @@
 <module>
 	<name>ps_googleanalytics</name>
 	<displayName><![CDATA[Google Analytics]]></displayName>
-	<version><![CDATA[4.2.2]]></version>
+	<version><![CDATA[4.3.0]]></version>
```
matks, line 5: If we drop old compatibility, maybe we should do it in a 5.0.0 ? So people understand it's a big change?

matks (commented):
Code is good for me 👍 I'm just not sure if we should do a 4.3.0 or a 5.0.0

tleon (approved):
I agree with @matks a drop in PS version support should be a major module version

config.xml
kpodemski: @Hlavtox it is actually **required** to bump the major version if we change PS compatibility version :)

kpodemski: https://github.com/PrestaShop/ADR/blob/master/0020-module-version-bump-after-changing-min-core-version.md

M0rgan01 (approved):

kpodemski (dismissed):
Version bump required

classes/Hook/HookDisplayBeforeBodyClosingTag.php
```diff
@@ -0,0 +1,152 @@
...
+                                ],
+                            ],
+                        ];
+                        $gaScripts .= $this->module->getTools()->renderEvent(
+                            'remove_from_cart',
+                            $eventData
+                        );
+                    }
+                } elseif (is_array($gacart)) {
+                    $it = new RecursiveIteratorIterator(new RecursiveArrayIterator($gacart));
```
matks, line 116: 😛

kpodemski (approved):

kpodemski (approved):

matks (approved):

Hlavtox:
@aniszr @hibatallahAouadni Issue fixed, you can test it :-)

nicosomb:
Thank you @Hlavtox !

PR #150, opened 2023-06-28
Fix remaining issues in order tracking

Hlavtox:
| Questions         | Answers
| Description?      | Fixes some issues with tracking in this module.
| Type?             | refactor
| BC breaks?        | yes
| Deprecations?     | no
| Fixed ticket?     | 
| Sponsor company   | 
| How to test?      | Test with https://tagassistant.google.com/ and see all events are working.

## Changes
- Unified BO & FO order wrapping, added item list to backoffice order.
- Payment info now using order currency not context.
- Purchase in BO now using order currency, not context.
- Added proper loading of manual orders created in backoffice.
- Added proper re-sending of failed orders for certain period. You can configure this and configure a maximal order age, so it doesn't load some ancient orders on first load, like the module used to do.

### Remaining issues with Google Analytics 4
- Go through everything again and check data structures.
- Implement item_category to item_category5 - https://github.com/PrestaShop/PrestaShop/issues/33022
- View_item_list not implemented
- View_cart not implemented


classes/Form/ConfigurationForm.php
```diff
@@ -112,7 +113,7 @@ public function generate()
                     'type' => 'switch',
                     'label' => $this->module->getTranslator()->trans('Anonymize IP', [], 'Modules.Googleanalytics.Admin'),
                     'name' => 'GA_ANONYMIZE_ENABLED',
-                    'hint' => $this->module->getTranslator()->trans('Use this option to anonymize the visitor’s IP to comply with data privacy laws in some countries', [], 'Modules.Googleanalytics.Admin'),
+                    'desc' => $this->module->getTranslator()->trans('Use this option to anonymize the visitor’s IP to comply with data privacy laws in some countries', [], 'Modules.Googleanalytics.Admin'),
```
leemyongpakvn, line 116: I think we can use `$this->module->trans` instead.

Hlavtox: @leemyongpakvn It didn't work for some reason, that's why I used `$this->module->getTranslator()`. 🤷‍♂️
See https://github.com/PrestaShop/ps_googleanalytics/pull/148#pullrequestreview-1474916493

micka-fdz (approved):
Everything seems OK to me.

classes/Database/Install.php
```diff
@@ -104,9 +108,30 @@ public function registerHooks()
...
+     *
+     * @return bool
+     */
+    public function installTab()
+    {
+        $tab = new Tab();
+        $tab->class_name = 'AdminGanalyticsAjax';
+        $tab->module = $this->module->name;
+        $tab->active = true;
+        $tab->id_parent = -1;
```
matks, line 129: Whoa 🤯 you're doing some nasty things here.

I understand you use this workaround to deal with a limitation of the software. Can you explain what is it? I'm curious if we can find a better cleaner solution for it

classes/GoogleAnalyticsTools.php
```diff
@@ -42,38 +42,39 @@ public function filter($gaScripts, $filterable)
...
-     * @param array $products
-     * @param array $transaction
+     * @param array $orderProducts
+     * @param array $orderData
+     * @param string $callbackUrl
      *
      * @return string|void
      */
-    public function addTransaction($products, $transaction)
+    public function renderPurchaseEvent($orderProducts, $orderData, $callbackUrl)
```
matks, line 53:
```suggestion
    public function renderPurchaseEvent(array $orderProducts, array $orderData, string $callbackUrl)
```
New function? Let's make it as clean as possible from day one 😉

classes/Hook/HookActionValidateOrder.php
```diff
@@ -0,0 +1,81 @@
...
+    {
+        // Check if we are creating backoffice order
+        if ($this->context->controller->controller_name != 'AdminOrders' && $this->context->controller->controller_name != 'Admin') {
+            return;
+        }
+
+        // Mark this ID to immediately display it on next page load
+        $order = $this->params['order'];
+        if (!empty($this->context->cookie->__get('ga_admin_order'))) {
+            $ga_admin_order = $this->context->cookie->__get('ga_admin_order') . ',' . $order->id;
```
matks, line 64: Suggestion: use sprintf instead.

I personnally find it easier to read ;) and I think it's less prone to errors.

classes/Hook/HookDisplayBackOfficeHeader.php
```diff
@@ -48,84 +51,139 @@ public function __construct(Ps_Googleanalytics $module, Context $context)
...
+
+        return $gaScripts;
+    }
+
+    /**
+     * Renders tracking code for given order
+     *
+     * @param int $idOrder
+     */
+    public function processOrder($idOrder)
```
matks, line 134: Same here ;) PHPDoc and descriptive function name.

In an ideal world we should be able to guess the function content just by reading the name

classes/Hook/HookDisplayOrderConfirmation.php
```diff
@@ -49,60 +49,58 @@ public function __construct(Ps_Googleanalytics $module, Context $context)
...
-                    [
-                        'id_order' => (int) $order->id,
-                        'id_shop' => (int) $this->context->shop->id,
-                        'sent' => 0,
-                        'date_add' => ['value' => 'NOW()', 'type' => 'sql'],
-                    ]
-                );
+        // Load up our handlers and repositories
+        $ganalyticsRepository = new GanalyticsRepository();
+        $gaTagHandler = new GanalyticsJsHandler($this->module, $this->context);
```
matks, line 61: You're loading services by yourself. Cannot we use (Symfony DI](https://symfony.com/doc/current/components/dependency_injection.html) instead ?

It will do the job nice and once for you ;)

classes/Repository/GanalyticsRepository.php
```diff
@@ -41,6 +41,21 @@ public function findGaOrderByOrderId($orderId)
...
     }
 
+    /**
+     * Checks if order is already sent to GA
+     *
+     * @param int $idOrder
+     *
+     * @return bool
+     */
+    public function orderAlreadySent($idOrder)
```
matks, line 51:
```suggestion
    public function hasOrderBeenAlreadySent($idOrder)
```

classes/Repository/GanalyticsRepository.php
```diff
@@ -78,6 +93,26 @@ public function addNewRow(array $data, $type = Db::INSERT_IGNORE)
...
+     * @return bool
+     */
+    public function addOrder(int $idOrder, int $idShop)
+    {
+        return $this->addNewRow(
+            [
+                'id_order' => (int) $idOrder,
+                'id_shop' => (int) $idShop,
+                'sent' => 0,
+                'date_add' => ['value' => 'NOW()', 'type' => 'sql'],
```
matks, line 111: Hoho ... I remember we used to have problems with such things. Possible warning.

@zuk3975 I remember we had a tough time because of something like this when working on a migrated page. Do you remember what it was ? 😅

controllers/admin/AdminGanalyticsAjax.php
```diff
@@ -30,14 +30,7 @@ public function init()
...
         if (Validate::isLoadedObject($order) && (isset($this->context->employee->id) && $this->context->employee->id)) {
-            (new GanalyticsRepository())->updateData(
-                [
-                    'sent' => 1,
-                    'date_add' => ['value' => 'NOW()', 'type' => 'sql'],
-                ],
-                'id_order = ' . $orderId
-            );
-
+            (new GanalyticsRepository())->markOrderAsSent((int) $orderId);
```
matks, line 33: 👍

matks (commented):
Code seems valid but I think it can be improved ;) see my suggestions

classes/Database/Install.php
kpodemski: @matks common pattern, I'm surprised it's the first time you see it

you want to have a tab to be available to perform ajax requests, but you hide it as it shouldn't be accessible normally

classes/GoogleAnalyticsTools.php
Hlavtox: We cant use named arguments because Google Analytics supports Prestashop versions with low PHP support. I think it's not worth to break the compatibility if there is no other benefit. 👍

classes/Hook/HookActionValidateOrder.php
```diff
@@ -0,0 +1,81 @@
...
+        $this->context->cookie->__set('ga_admin_order', $ga_admin_order);
+        $this->context->cookie->write();
+    }
+
+    /**
+     * setParams
+     *
+     * @param array $params
+     */
+    public function setParams($params)
```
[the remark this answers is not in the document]
Hlavtox: It's like this for all other hooks, so for sake of consistency. :-)

classes/Hook/HookActionValidateOrder.php
Hlavtox: Will do, using dots is me still being a beginner. :-)

classes/Database/Install.php
Hlavtox: > Almost every month I find something in prestashop that I never knew before XD it's so big

I didn't know also until last month Mathieu don't worry. 🤣 It seems like a hackjob but this is how it's done. You need to have every backoffice controller registered as a tab, or it won't find it and it throws error that controller not found.

classes/Hook/HookActionValidateOrder.php
kpodemski: @Hlavtox you can also stop using `__get` or `__set`, you can just do 

`$this->context->cookie->ga_admin_order` which calls `__get`

or `$this->context->cookie->ga_admin_order = 'something'` which calls `__set` 👍🏻

classes/Hook/HookActionValidateOrder.php
```diff
@@ -0,0 +1,81 @@
...
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\Module\Ps_Googleanalytics\Hooks;
+
+use Context;
+use Ps_Googleanalytics;
+
+class HookActionValidateOrder implements HookInterface
```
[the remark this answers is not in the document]
Hlavtox: Same here - PHP compatibility.

classes/Repository/GanalyticsRepository.php
Hlavtox: Just moved the code 👍

classes/Hook/HookActionValidateOrder.php
Hlavtox: @kpodemski Fixed

classes/Hook/HookDisplayHeader.php
```diff
@@ -120,14 +118,6 @@ private function displayGaTag()
...
-        $this->acknowledgeBackOfficeContext($backOffice);
-    }
-
-    /**
-     * @param bool $isBackOffice
-     */
-    public function acknowledgeBackOfficeContext($isBackOffice)
-    {
-        $this->backOffice = $isBackOffice;
+        $this->backOffice = $backOffice;
```
leemyongpakvn, line 121: @matks disagreed with [this kind of change last year](https://github.com/PrestaShop/ps_googleanalytics/pull/122/files). Is it OK this time ? :stuck_out_tongue_winking_eye:

Hlavtox: This version of module will break lot of things, so we can finally remove this nonsense. :-))

matks: @leemyongpakvn In the PR https://github.com/PrestaShop/ps_googleanalytics/pull/122 the BC break was alone, so merging it would have required to release a major version just for one little BC break. I felt like it was not a good balance.

If we aim for a bigger major release, packed with BC breaks, I think we can accept it.

It's like going to the supermarket with your car 😄 if you tell me "I just want to buy a bottle of water" I feel like it's not worth it (using the car, using gas, spending time driving... all this just for one bottle) ; if you tell me "I'm doing the groceries for the two next weeks" then it makes sense

M0rgan01 (approved):
thanks for the comments, they make the review easier

classes/Hook/HookActionValidateOrder.php
```diff
@@ -0,0 +1,87 @@
...
+
+    /**
+     * run
+     *
+     * @return void
+     */
+    public function run()
+    {
+        // Check if we are creating backoffice order
+        if ($this->context->controller->controller_name != 'AdminOrders' && $this->context->controller->controller_name != 'Admin') {
```
kpodemski, line 57: @Hlavtox 

hey Daniel

you sure you wanted to do `$this->context->controller->controller_name != 'Admin'`? when's situation where the controller is `Admin`? 🤔 maybe you wanted to use `strpos` there?

Hlavtox: Can't remember after 3 months, I need to check it again

Hlavtox: @kpodemski Yes it's correct.

`actionValidateOrder` - `dump($this->context->controller);`

Dies on this URL: http://localhost/81x/admin-dev/index.php/sell/orders/place?_token=dkwJoC-fWxJi0Qr5YgeSYtTvGmT1OOPEi4JbJqqDbjQ

Snímek obrazovky 2023-10-30 172347

kpodemski (approved):
After discussing with @Hlavtox I'm merging this PR. The work on version 5 of Google Analytics is still in progress, but this PR unblocks some of the future tasks. The full tests will be performed before the release of the module.

PR #153, opened 2023-11-03
Implement remaining tracking events for GA4

Hlavtox:
| Questions         | Answers
| Description?      | I completely refactored the modules and implemented all events I could implement. All data is standardized through events with no inconsistencies. Removed all old things.
| Type?             | refactor
| BC breaks?        | yes
| Deprecations?     | no
| Fixed ticket?     | https://github.com/PrestaShop/PrestaShop/issues/33022, https://github.com/PrestaShop/PrestaShop/issues/31931
| Sponsor company   | 
| How to test?      | Use https://tagassistant.google.com/, put your shop there and you will see all events that are being echoed.

### Description
- Implemented view_item event on product visits.
- Implemented select_item event if user is coming from our shop, if the user came from a product listing, we will also provide this information.
- Implemented select_item for quickview clicks.
- Implemented view_item_list event for all listings, with proper names.
- Implemented search event.
- Implemented view_cart event.
- Implemented begin_checkout event for first visits to checkout. Checked by a referrer.
- Implemented sign_up event.
- Implemented login event.
- Implemented cart adding/increasing/decreasing/removing.
- Unified product data everywhere, including variant information.

### Before you ask/known limitations
- I use different hooks for cart events because they have much better information about the action done and they provide more information. With actionCartSave, we only guessed if the actions will succeed, and for example, it didn't provide information about what variant was being added.
- Cart adding/removing events, delivery and payment info events are rendered into data repository and flushed on first chance, on a next page load. It would be good to make it dynamic by javascript, but I didn't want to touch that. :-)
- Listing renders only on first page load, if you use pagination and the page changes dynamically, it's not refreshed. We would need to dynamically render these parts of the page somehow. The same goes for quickview events.


classes/Hook/HookActionCartUpdateQuantityBefore.php
```diff
@@ -0,0 +1,107 @@
...
+    private $params;
+
+    public function __construct(Ps_Googleanalytics $module, Context $context)
+    {
+        $this->module = $module;
+        $this->context = $context;
+    }
+
+    /**
+     * run
```
matks, line 45: Not very useful 😛

classes/Hook/HookDisplayBeforeBodyClosingTag.php
```diff
@@ -49,100 +44,236 @@ public function __construct(Ps_Googleanalytics $module, Context $context)
...
+        $items = $productWrapper->prepareItemListFromProductList($listing['products']);
+
+        // Prepare info about the list
+        $item_list_id = $this->context->controller->php_self;
+        $item_list_name = $listing['label'];
 
-        if ($controller_name == 'order' || $controller_name == 'orderopc') {
+        // Render the event
+        $eventData = [
+            'item_list_id' => $item_list_id,
```
matks, line 89: Is it OK to use `$this->context->controller->php_self` as ID?

classes/Hook/HookDisplayBeforeBodyClosingTag.php
```diff
@@ -49,100 +44,236 @@ public function __construct(Ps_Googleanalytics $module, Context $context)
...
+            'item_list_name' => $item_list_name,
+            'items' => $items,
+        ];
+        $this->gaScripts .= $this->module->getTools()->renderEvent(
```
matks, line 93: Since you are doing a refactoring, can I suggest we stop the "append something here ... something here ..." logic?

We could do this:
- create an array / an object `$GAScriptParts`
- in the code, we slowly fill in this object
- at the end we compute the final string

Easy implementation: just concatenate each part with an `implode()`
Better implementation: smarter logic similar to what a templating engine would do (like smarty or twig)

classes/Hook/HookDisplayBeforeBodyClosingTag.php
```diff
@@ -49,100 +44,236 @@ public function __construct(Ps_Googleanalytics $module, Context $context)
...
-                'begin_checkout',
-                $eventData
+
+            // Keep only product ID if id_product_attribute was appended
+            $productId = explode('-', $item['item_id']);
+            $productId = $productId[0];
+
+            // Render the event wrapped in onclick
+            $this->gaScripts .= '
+            $(\'article[data-id-product="' . $productId . '"] a.quick-view\').on(
```
matks, line 112: 🤮 
Brrrrr 😰 

That's why I suggest to better wrap the string computation part 😅

classes/Hook/HookDisplayFooterProduct.php
```diff
@@ -42,73 +39,107 @@ public function __construct(Ps_Googleanalytics $module, Context $context)
...
-        if (isset($_SERVER['HTTP_REFERER']) && strpos($_SERVER['HTTP_REFERER'], $_SERVER['HTTP_HOST']) > 0) {
-            $js .= $this->module->getTools()->addProductClickByHttpReferal([$gaProduct], $this->context->currency->iso_code);
+        // If the user got to the product page from previous page on our shop,
+        // we will also send select_item event.
+        if ($this->wasPreviousPageOurShop()) {
+            $eventData = [
+                'items' => [$item],
+            ];
+
+            // We will also try to get the information about the last visited listing.
```
matks, line 84: So many rules "if we are on page (...) then we do (...) but if we come from page (...) then we do (...)" it's hard to build a complete mental map of all the usecases.

I wonder if all this logic is in the right place in HookDisplayFooterProduct. How could we make it easier to know "if I am on hook (...) and doing (...) then data (...) will be sent" ?

classes/Wrapper/OrderWrapper.php
```diff
@@ -44,14 +43,14 @@ public function wrapOrder($order)
...
+            'transaction_id' => (int) $order->id,
             'affiliation' => Shop::isFeatureActive() ? $this->context->shop->name : Configuration::get('PS_SHOP_NAME'),
-            'revenue' => (float) $order->total_paid,
+            'value' => (float) $order->total_paid,
             'shipping' => (float) $order->total_shipping,
             'tax' => (float) $order->total_paid_tax_incl - $order->total_paid_tax_excl,
             'customer' => (int) $order->id_customer,
             'currency' => $currency->iso_code,
-            'payment_type' => (int) $order->payment,
+            'payment_type' => (string) $order->payment,
```
matks, line 53: This change (and others) make me thing: this is going to make a clear difference in data being recorded with ps_googleanalytics modules 4 and 5, right? We need to make people fully aware of that. Because they won't be able to compare data collected and data collected after?

classes/Wrapper/ProductWrapper.php
```diff
@@ -20,97 +20,237 @@
...
+                $productIds[] = $product['id'];
             }
+        }
+        $this->loadCategories($productIds);
 
-            if (!isset($product['price'])) {
-                $product['price'] = (float) Tools::displayPrice(Product::getPriceStatic((int) $product['id_product'], $usetax), $currency);
-            }
-            $result_products[] = $this->wrapProduct($product, [], $index);
+        // Prepare each item and override the counter
```
matks, line 70: I'm not sure to understand why you need to override the counter

classes/Wrapper/ProductWrapper.php
```diff
@@ -20,97 +20,237 @@
...
+        $counter = 1;
+        foreach ($productCategories as $productCategory) {
+            // Skip home category if we have more than 1 category
+            if (count($productCategories) > 1 && $productCategory['id_category'] == $this->homeCategory) {
+                continue;
+            }
+
+            // Add it with proper key
+            $item[$counter == 1 ? 'item_category' : 'item_category' . $counter] = $productCategory['name'];
+            if ($counter == 5) {
```
matks, line 214: Again some rules 😄 that are hard to understand. What is special with counter 5?

Sorry for asking all these questions, but I'm sure the next developer who tries to read the code will ask the same 😉

classes/Wrapper/OrderWrapper.php
Hlavtox: This was a bug

classes/Hook/HookDisplayBeforeBodyClosingTag.php
Hlavtox: Yes, it's a clear identifier used in all modules.

classes/Wrapper/ProductWrapper.php
Hlavtox: https://developers.google.com/analytics/devguides/collection/ga4/reference/events?client_type=gtag#view_item

Because we can have maximum of item_category5 in the event.

classes/Wrapper/ProductWrapper.php
Hlavtox: Because I render `index: 0` when doing a single product.
When rendering a list, I will do 0,1,2... etc.

classes/Hook/HookDisplayFooterProduct.php
Hlavtox: Yeah, but that's just how it's done, its million times clearer than the old code. :-)

Hlavtox:
Merging after agreeing with QA team, we will do full test on release.

PR #159, opened 2024-02-20
Update default granted consents

Hlavtox:
| Questions         | Answers
| Description?      | Adds default granted consents to make sure all data is allowed to flow into GA.
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | https://support.google.com/tagmanager/answer/13695607
| Sponsor company   | 
| How to test?      |


kpodemski (changes requested):
It cannot work like that. It should be denied by default, and granted when user consents.

It is not responsibility of the Analytics module to handle consent/cookie banners fully.

Hlavtox:
PR not needed

PR #161, opened 2024-02-28
Check if property exists on controller

Hlavtox:
| Questions         | Answers
| Description?      | controller_name property was throwing some notices, we check it before accessing it.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     |no
| Fixed ticket?     | 
| Sponsor company   | 
| How to test?      |

### DIFF
--- a/classes/Hook/HookActionValidateOrder.php
+++ b/classes/Hook/HookActionValidateOrder.php
@@ -53,8 +53,9 @@ class HookActionValidateOrder implements HookInterface
      */
     public function run()
     {
-        // Check if we are creating backoffice order
-        if ($this->context->controller->controller_name != 'AdminOrders' && $this->context->controller->controller_name != 'Admin') {
+        // Check if we are creating backoffice order, we are only launching this hook when creating backoffice order
+        // For FO purposes, we use displayOrderConfirmation.
+        if (empty($this->context->controller->controller_name) || ($this->context->controller->controller_name != 'AdminOrders' && $this->context->controller->controller_name != 'Admin')) {
             return;
         }
 


classes/Hook/HookActionValidateOrder.php
```diff
@@ -53,8 +53,9 @@ public function __construct(Ps_Googleanalytics $module, Context $context)
      */
     public function run()
     {
-        // Check if we are creating backoffice order
-        if ($this->context->controller->controller_name != 'AdminOrders' && $this->context->controller->controller_name != 'Admin') {
+        // Check if we are creating backoffice order, we are only launching this hook when creating backoffice order
+        // For FO purposes, we use displayOrderConfirmation.
+        if (empty($this->context->controller->controller_name) || ($this->context->controller->controller_name != 'AdminOrders' && $this->context->controller->controller_name != 'Admin')) {
```
matthieu-rolland, line 58: maybe this is clearer ?

```suggestion
        if (isset($this->context->controller->controller_name) &&  !in_array($this->context->controller->controller_name, ['Admin', 'AdminOrders']) {
```

matthieu-rolland (approved):

classes/Hook/HookActionValidateOrder.php
jolelievre, line 58:
```suggestion
        if (empty($this->context->controller) || ($this->context->controller->controller_name != 'AdminOrders' && $this->context->controller->controller_name != 'Admin')) {
```

jolelievre: The bug in behat is because `Content::controller` is not set it is null so accessible `null->controller_name` will still fail with this fix no? 🤔

matthieu-rolland: I had a doubt to but I think `empty()` is very forgiving, `empty` will evaluate to true at the first unset property

Hlavtox: @jolelievre The issue was in controller_name, the fix works correctly.

PR #164, opened 2024-03-04
Fix bug breaking price cache keys used in getPriceStatic and priceCalculation

matthieu-rolland:
| Questions         | Answers
| Description?      | The method `getProductProperties` was sent a product without an `id_product_attribute` specified, which caused a price cache key to not be build properly
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | 
| Sponsor company   | PrestaShop SA
| How to test?      |

### DIFF
--- a/classes/Hook/HookActionCartUpdateQuantityBefore.php
+++ b/classes/Hook/HookActionCartUpdateQuantityBefore.php
@@ -66,15 +66,13 @@ class HookActionCartUpdateQuantityBefore implements HookInterface
         // Format product and standardize ID
         $product = (array) $this->params['product'];
         $product['id_product'] = $product['id'];
-        $product['id_product_attribute'] = $this->params['id_product_attribute'];
-
-        // Get some basic information
-        $product = Product::getProductProperties($this->context->language->id, $product);
-
         // Add information about attribute
         if (!empty($this->params['id_product_attribute'])) {
             $product['id_product_attribute'] = (int) $this->params['id_product_attribute'];
         }
+
+        // Get some basic information
+        $product = Product::getProductProperties($this->context->language->id, $product);
 
         // Add information about quantity difference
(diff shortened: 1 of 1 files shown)


kpodemski (approved):
Nice catch!

Hlavtox (changes requested):
Sorry but I don't see how is this a solution?

A) You can check few lines below that I send the parameter if it's available.

```php
// Add information about attribute
        if (!empty($this->params['id_product_attribute'])) {
            $product['id_product_attribute'] = (int) $this->params['id_product_attribute'];
        }
```

B) You can check in Product::getProductProperties, that sending no `id_product_attribute` is a perfectly valid syntax and it will automatically retrieve default attribute inside using `Product::getDefaultAttribute`.

matthieu-rolland:
> Sorry but I don't see how is this a solution?
> 
> A) You can check few lines below that I send the parameter if it's available.
> 
> ```
> // Add information about attribute
>         if (!empty($this->params['id_product_attribute'])) {
>             $product['id_product_attribute'] = (int) $this->params['id_product_attribute'];
>         }
> ```
> 
> B) You can check in Product::getProductProperties, that sending no `id_product_attribute` is a perfectly valid syntax and it will automatically retrieve default attribute inside using `Product::getDefaultAttribute`.

It had to be done before the call of `getProductProperties`, `getPriceStatic` and `priceCalculation` methods need this information.

Hlavtox (approved):

kpodemski (approved):

Hlavtox:
After internal discussion - the fix is legit because it provides a proper combination ID into the logic and provides a correct price for a product.

But the issue with failing tests lies inside `Product::getProperties` method, which returns `null` for a default combination in test context, for some reason.

PR #167, opened 2024-05-14
Confirm if

fulldecent:
Please confirm if 8x. is supported so I can add to https://github.com/fulldecent/awesome-prestashop-modules


matks (commented):
Hello @fulldecent yes 8.x is supported however I am not convinced this change is a good idea

Else, we'll have to come again on this README when we release PrestaShop v9, and PrestaShop v10, and ... 😄 

The "later" means "any upper version" 😉

PR #168, opened 2024-06-16
Update ps_googleanalytics.php

panariga:
wrap hooks code in try...catch to avoid site crashes

| Questions         | Answers
| Description?      |Let's wrap the hooks code in a try...catch block. If any error occurs, the Analytics module may crash and cause a 500 error, which is unacceptable for the front-end. For instance, crashes can occur if paths saved in a cookie are too long, exceeding the maximum allowed size, or if non-ASCII paths have improperly escaped characters. It's better to lose some analytics events than to lose customers due to errors.
| Type?             | bug fix / improvement 
| BC breaks?        | no
| Deprecations?     |  no
| Fixed ticket?     | no
| Sponsor company   | no
| How to test?      | additional test not needed

### DIFF
--- a/ps_googleanalytics.php
+++ b/ps_googleanalytics.php
@@ -80,10 +80,14 @@ class Ps_Googleanalytics extends Module
 
     public function hookDisplayHeader($params, $back_office = false)
     {
-        $hook = new PrestaShop\Module\Ps_Googleanalytics\Hooks\HookDisplayHeader($this, $this->context);
-        $hook->setBackOffice($back_office);
+        try {
+            $hook = new PrestaShop\Module\Ps_Googleanalytics\Hooks\HookDisplayHeader($this, $this->context);
+            $hook->setBackOffice($back_office);
 
-        return $hook->run();
+            return $hook->run();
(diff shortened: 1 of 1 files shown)


Hlavtox (changes requested):
Sorry @panariga, but this is not the proper way to do things. 👍 If there is some issue, it must be fixed.

panariga:
I agree that as many issues as can be fixed should be fixed. However, some can't be fixed by design. For example, Bingbot tries all possible variations for faceted search combinations. This results in super-long queries and the module tries to save them to a cookie. Because of that, the operation fails due to the cookie size limitations. This requires hard refactoring to create new storage.

But in general, this module is just for stats. It shouldn’t break the shop's work. Due to the complicated information-gathering routines, you can never guarantee an error-free flow. So, anyway, wrapping in try..catch has no regressions—you still have errors in the console, but users are not frustrated by these errors.

Hlavtox:
@panariga Well, this is actually quite a legit issue we could definitely fix, either by storing only the url without a query string, or by using a storage for this.

PR #178, opened 2026-06-02
CI: align PHPStan on PS 8.2+, refresh neon configs and fix renderCartPage error

mattgoud:
| Questions         | Answers
| ----------------- | -------------------------------------------------------
| Branch?           | dev
| Description?      | The `phpstan-80-84` CI job targeted PrestaShop `9.0.x` which no longer exists as a branch, causing systematic PHPStan failures. Rather than dropping PS 9.0 coverage, this PR moves it into the main `phpstan` job using the `9.0.3` tag instead of the removed branch.<br><br>This PR also aligns the module on the PS 8.2+ baseline targeted by the cleanup Epic:<br>- drops the `8.1.7` PHPStan target (the PHP 7.4/8.1 job now runs `8.2.x` only)<br>- replaces the dead `phpstan-9.0.x.neon` (branch) with `phpstan-9.0.3.neon` (tag) and removes the orphaned `phpstan-8.1.7.neon`<br>- the `phpstan` job now targets `9.0.3`, `9.1.x` and `develop` (PHP 8.1–8.5, with `9.0.3` capped at PHP 8.4 since PS 9.0.x does not support 8.5)<br>- expands `php-linter` to check every PHP version from 7.2 to 8.5 (the `phpstan-74` analysis job stays on PHP 7.4, as the module's dev tooling — `prestashop/php-dev-tools` — requires it)<br>- bumps `ps_versions_compliancy` min to `8.2.0`<br>- removes 4 orphaned legacy PHPStan neon configs no longer referenced by the matrix (`1.7.7`, `1.7.8`, `8.0`, `latest`)<br>- fixes a PHPStan error directly instead of suppressing it: aligns the `renderCartPage()` method casing with its call site (PHP is case-insensitive so it worked at runtime, but the PHPStan 0.12 job flagged it as unused). The remaining `ignoreErrors` are documented inline as genuine framework/structural limitations: Cookie magic `__get`/`__set` properties, `parent::uninstall()` called through a closure rebound via `Closure::bindTo()`, and the `HookActionValidateOrder::$module` property (the `HookInterface` constructor signature mandates the `$module` argument, so it must be stored even in hooks that never read it back).
| BC breaks?        | no
| Deprecations?     | no
| Fixed issue or discussion?     | Part of PrestaShop/PrestaShop#41648
| Sponsor company   | PrestaShop SA


ps-jarvis:
Hello @mattgoud!

This is your first pull request on [ps_googleanalytics](https://github.com/PrestaShop/ps_googleanalytics/) repository of the PrestaShop project.

Thank you, and welcome to this Open Source community!

.github/workflows/php.yml
```diff
@@ -66,40 +66,6 @@ jobs:
           phpstan-version: '^0.12'
           composer-version: '2.2.18'
 
-  # Run PHPStan against the module (PHP 8.1 – 8.4)
-  phpstan-80-84:
```
jolelievre, line 70: Why remove those?

mattgoud: ok @jolelievre good point, re-added it using the `9.0.3` tag instead of the deleted `9.0.x` branch: `presta_version: ['9.0.3', '9.1.x', 'develop']` + matching `phpstan-9.0.3.neon`.

jolelievre (approved):

.github/workflows/php.yml
jf-viguier: According to source of truth here https://github.com/PrestaShop/PrestaShop/issues/41648
9.0.3 is not supported, only 9.1.x and develop

.github/workflows/php.yml
```diff
@@ -66,48 +81,21 @@ jobs:
...
   phpstan:
     name: PHPStan (PHP 8.1 - 8.5)
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        presta_version: ['9.1.x', 'develop']
-        php_version: ['8.1', '8.5']
+        include:
+          # PrestaShop 9.0.x supports PHP 8.1 up to 8.4 (8.5 is not supported)
+          - { presta_version: '9.0.3', php_version: '8.1' }
```
jf-viguier, line 92: Please use matrix style 

```
presta_version: ['9.1.x', 'develop']
php_version: ['8.1', '8.5']
```

jf-viguier (dismissed):
Small changes asked to fit source of truth from  @Progi1984 https://github.com/PrestaShop/PrestaShop/issues/41648

.github/workflows/php.yml
mattgoud: hello @jf-viguier, We need to test tag **9.0.3** (because the 9.0.x branch no longer exists). I've forwarded your comment to the team to be sure. The goal is to keep the modules up-to-date **without any gaps**, because even shops on **9.0.x** may need to upgrade their modules. Therefore, it was decided to adopt and maintain a compatibility range for the modules.

mattgoud: @jf-viguier, for the matrix, we can't use the previous style because, as we're testing **9.0.3**, this version doesn't support PHP 8.5 (hence the need for a more atomic approach using the "includes" attribute).

mattgoud:
correction: this is a major bump, not a patch — version set to 6.0.0 (ps_googleanalytics.php + config.xml)

jolelievre (approved):

ps-jarvis:
PR merged, well done!

Message to @PrestaShop/committers: do not forget to milestone it before the merge.

PR #181, opened 2026-06-27
Prevent fatal error when product template variable is a Product object

### LINKED ISSUE #179: Fatal error
### Prerequisites

- [x] I understand and accept the project's [code of conduct](https://github.com/PrestaShop/PrestaShop/blob/develop/CODE_OF_CONDUCT.md).
- [x] I have already [searched in existing issues](./issues?q=is%3Aissue+label%3ABug) and found no previous report of this bug.

### Describe the bug and add attachments

Error appears in product page

Fatal error: Uncaught Error: Cannot use object of type Product as array in /home/zoshop/domains/zzzzzzzz.lt/public_html/modules/ps_googleanalytics/classes/Wrapper/ProductWrapper.php:136 Stack trace: #0 /home/zoshop/domains/zzzzzzzz.lt/public_html/modules/ps_googleanalytics/classes/Hook/HookDisplayFooterProduct.php(62): PrestaShop\Module\Ps_Googleanalytics\Wrapper\ProductWrapper->prepareItemFromProduct() #1 /home/zoshop/domains/zzzzzzzz.ltt/public_html/modules/ps_googleanalytics/ps_googleanalytics.php(120): PrestaShop\Module\Ps_Googleanalytics\Hooks\HookDisplayFooterProduct->run() #2 /home/zoshop/domains/zzzzzzzz.lt/public_html/classes/Hook.php(1083): Ps_Googleanalytics->hookDisplayFooterProduct() #3 /home/zoshop/domains/zzzzzzzz.lt/public_html/classes/Hook.php(421): HookCore::coreCallHook() #4 /home/zoshop/domains/zzzzzzzz.lt/public_html/classes/Hook.php(989): HookCore::callHookOn() #5 /home/zoshop/domains/zzzzzzzz.lt/public_html/config/smarty.config.inc.php(201): HookCore::exec() #6 /home/zoshop/domains/zzzzzzzz.lt/public in /home/zoshop/domains/zzzzzzzz.lt/public_html/modules/ps_googleanalytics/classes/Wrapper/ProductWrapper.php on line 136

### Steps to reproduce

After update to 5.0.3

### Expected behavior

_No response_

### Actual Result

_No response_

### PrestaShop version where the bug happens

8.2.4 

### Module version where the bug happens

5.0.3

### PHP version(s) where the bug happened

_No response_

### Your company or customer's name goes here (if applicable).

_No response_

boo-code:
| Questions         | Answers
| ----------------- | -------------------------------------------------------
| Description?      | `HookDisplayFooterProduct::run()` reads the `product` Smarty variable and passes it straight to `ProductWrapper::prepareItemFromProduct()`, which consumes it with array keys (`$product['id_product']`, `$product['name']`, …). That variable is normally a `ProductLazyArray` (array-accessible), but some themes/modules assign a **raw `Product` object** to it, producing a fatal error `Cannot use object of type Product as array` on the product page (reported on 5.0.3 / PS 8.2.4). This PR guards the hook: if the product is not array-accessible, it skips tracking instead of breaking the page. `ProductLazyArray` implements `ArrayAccess`, so the normal flow is unaffected.
| Type?             | bug fix
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | Fixes PrestaShop/ps_googleanalytics#179.
| How to test?      | On a product page where the `product` Smarty variable is a raw `Product` object, before this PR the page fatals with `Cannot use object of type Product as array` (`ProductWrapper.php` line ~136). Deterministic check: `prepareItemFromProduct(new Product(1))` throws that error; with the guard, `HookDisplayFooterProduct::run()` returns without a fatal. With the standard `ProductLazyArray` (which implements `ArrayAccess`), tracking continues to work as before.
| Sponsor company   |

### DIFF
--- a/classes/Hook/HookDisplayFooterProduct.php
+++ b/classes/Hook/HookDisplayFooterProduct.php
@@ -54,6 +54,14 @@ class HookDisplayFooterProduct implements HookInterface
             return;
         }
 
+        // The 'product' template variable is expected to be a ProductLazyArray (array-accessible).
+        // Some themes or modules assign a raw Product object instead, which prepareItemFromProduct()
+        // cannot consume (it accesses array keys) and would raise a fatal error on the product page.
+        // Skip tracking in that case rather than breaking the page.
+        if (!is_array($product) && !$product instanceof \ArrayAccess) {
+            return;
+        }
+
         // Initialize tag handler
         $gaTagHandler = new GanalyticsJsHandler($this->module, $this->context);
 


ps-jarvis:
Hello @boo-code!

This is your first pull request on [ps_googleanalytics](https://github.com/PrestaShop/ps_googleanalytics/) repository of the PrestaShop project.

Thank you, and welcome to this Open Source community!

mattgoud (approved):

Hlavtox (commented):
@boo-code I am don't have a great feeling from this change. I think we should not add checks for bad modules overwriting system variables..

Cc @kpodemski @Codencode

kpodemski (approved):
To me the change is totally fine, and I'm always recommending to check whether the input data is what we expect.

boo-code:
Thanks @Hlavtox, that is a fair principle and I do not want this to become a pattern of guarding against every misbehaving module either. Let me explain why I think this specific case is worth it, and offer a middle ground.

The distinction I would draw: `product` here is not this module's own state, it is a shared Smarty template variable that the hook reads but does not set. It is populated by the theme, core, or another module, so from this hook's point of view its type is genuinely external input crossing a trust boundary, not a value we control. The guard validates input at that boundary rather than compensating for one known-buggy module.

The failure modes are also very asymmetric. Without the guard, a third party assigning a raw `Product` object produces an uncaught fatal that takes down the whole product page (issue #179 is exactly this, on a live shop), not just GA tracking. With the guard, the only consequence is that GA tracking is skipped for that one request. Degrading the analytics is proportionate; crashing the storefront is not.

It is also deliberately minimal and does not touch the normal path: `ProductLazyArray` implements `ArrayAccess`, so the standard flow is unaffected and only the anomalous type returns early.

On your underlying concern, that we should not silently mask a bad module: I agree, and I am happy to add a debug-level log when the guard trips, so the misconfiguration stays visible and diagnosable instead of being swallowed. If you would prefer that, I will push it; otherwise I am glad to leave the guard as it is. Either way, thanks for the careful read.

Codencode:
Indeed, I also think it's better to avoid the page crashing.

Codencode (approved):

